PR #22, opened 2016-06-22
Request is not needed to define policy

clelland:
The *Process response policy* algorithm specifies a `request` as input, but does not reference it in any way.


igrigorik:
👍

PR #33, opened 2016-08-15
[WIP] Use IDL to control if feature is exposed

igrigorik:
Preview: https://cdn.rawgit.com/WICG/feature-policy/idl/index.html#features

### DIFF
--- a/index.html
+++ b/index.html
@@ -453,6 +453,15 @@
           </tr>
           <tr>
             <td>
+              <a>payment</a>
+            </td>
+            <td>`self`</td>
+            <td>`null`</td>
+            <td>`null`</td>
+            <td>Controls access to [PaymentRequest interface].</td>
+          </tr>
+          <tr>
+            <td>
               <a>usermedia</a>
             </td>
             <td>`self`</td>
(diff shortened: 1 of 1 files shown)


igrigorik:
@raymeskhoury @clelland ptal. any red flags so far?

clelland:
I think this is looking good -- I'm still thinking about the defaults matrix, but it's making more sense.

Also, you should be able to apply the `[Feature]` IDL attribute to an entire partial interface if you want to, rather than just the individual members; it doesn't make much difference here, but if there were a larger interface then it would make it clear that the whole thing is a single 'feature' for FP purposes.

stefhak:
This mostly looks good to me. I don't remember us discussing controlling the access to webrtc in iframes, but I guess it can't hurt to have that (default is enabled).

Not related to usermedia and webrtc: at first glance I did not see how a chain of embeddings would work. I mean, if the top-level allows feature X for https://foo.com and https://bar.com, an iframe controlled by bar.com could enable or disable feature X for iframes from https://bar.com, but never enable it for other domains?

igrigorik:
@stefhak I haven't committed the enforcement algorithms yet (soon, stay tuned). You can see the sketch and the related discussion here: https://github.com/WICG/feature-policy/issues/28

foolip:
Looks good to me, but I'l leave to to @alvestrand to say if the token "webrtc" is appropriate for hiding the RTCPeerConnection interface.

igrigorik:
Merged. However, note that this does not (yet) reflect the latest discussions around migrating to the whitelist-only model... we'll tackle that once we have consensus on it via separate CLs.

PR #54, opened 2017-02-22
Major update to spec; rewrite in bikeshed.

clelland:
This is a significant update to the Feature Policy spec; in addition to
rewriting the spec in Bikeshed, this brings it in line with the explainer
document, adds iframe policies, removes explicit references to JFV, moves all
of the features into an appendix, (adding a couple of new features, and removing
references to features that we're not planning on trying to standardize on and
ship in the near future.)

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1,0 +1,1016 @@
+<pre class="metadata">
+Title: Feature Policy
+Shortname: feature-policy
+Level: 1
+Status: CG-DRAFT
+Group: WICG
+URL: https://wicg.github.io/feature-policy/
+Editor: Ilya Grigorik, Google, igrigorik@google.com
+Editor: Mike West, Google, mkwst@google.com
+Abstract: This specification defines a mechanism that allows developers to selectively enable and disable use of various browser features and APIs.
+Repository: https://github.com/wicg/feature-policy/
+Markup Shorthands: css no, markdown yes
+</pre>
+<pre class="link-defaults">
+spec:html; type:interface; for:/; text:Document
+spec:url; type:dfn; text:origin
+spec:fetch; type:dfn; for:Response; text:response
+spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:element; text:script
+spec:html; type:element; text:link
+spec:html; type:dfn; for:/; text:global object
+spec:html; type:dfn; for:/; text:header list
+spec:fetch; type:dfn; text:name
+spec:fetch; type:dfn; text:value
+</pre>
+<section>
+  <h2 id="introduction">Introduction</h2>
+  <p>The web-platform provides an ever-expanding set of features and APIs,
+  offering richer functionality, better developer ergonomics, and improved
+  performance. However, a missing piece is the ability for the developer to
+  selectively enable, disable, or modify the behavior of some of these browser
+  features and APIs within their application:</p>
+  <ol>
+    <li>The developer may want to selectively *disable* access to certain
+    browser features and APIs to "lock down" their application, as a security
+    or performance precaution, to prevent own and third-party content executing
+    within their application from introducing unwanted or unexpected behaviors
+    within their application.</li>
+    <li>The developer may want to selectively *enable* access to certain
+    browser features and APIs which may be disabled by default - e.g. some
+    features may be disabled by default in embedded context unless explicitly
+    enabled; some features may be subject to other policy requirements.</li>
+    <li>The developer may want to use the policy to assert a promise to a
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -0,0 +1,1016 @@
...
+        of <code>allow="fullscreen"</code> will be used.
+      </div>
+    </section>
+    <section>
+      <h4 id="iframe-allowpaymentrequest-attribute">allowpaymentrequest</h4>
+      <p>The "<code>allowpaymentrequest</code>" iframe attribute controls
+      access to [=Payment interface=].</p>
+      <p>If the iframe element has an "<code>allow</code>" attribute whose
+      value contains the token "<code>payment</code>", then the
+      "<code>allowpaymentrequest</code> attribute should have no effect.</p>
```
zcorpan: Why not must?

index.bs
```diff
@@ -0,0 +1,1016 @@
...
+    <p>Some features controlled by Feature Policy have existing iframe
+    attributes defined. This specification redefines these attributes to act as
+    declared policies for the iframe element.</p>
+    <section>
+      <h4 id="iframe-allowfullscreen-attribute">allowfullscreen</h4>
+      <p>The "<code>allowfullscreen</code>" iframe attribute controls access to
+      {{requestFullscreen()}}.</p>
+      <p>If the iframe element has an "<code>allow</code>" attribute whose
+      value contains the token "<code>fullscreen</code>", then the
+      "<code>allowfullscreen</code> attribute should have no effect.</p>
```
zcorpan: Why not must?

index.bs
```diff
@@ -0,0 +1,1016 @@
...
+      <div class="note">
+        This section extends HTML's <code>meta</code> [pragma directives].
+      </div><strong>Feature policy state (http-equiv="feature-policy")</strong>
+      <p>This pragma extends the declared feature policy for a {{Document}}.</p>
+      <ol>
+        <li>If the [meta] element has no [content] attribute, or if that
+        attribute's value is the empty string, then abort these steps.</li>
+        <li>If the [meta] element is not being inserted in into the
+        {{Document}}'s [head] element, or if is inserted after any [script] or
+        [link] elements, then abort these steps.</li>
```
zcorpan: Why this rule? Is "after" in time, or in tree order?

index.bs
```diff
@@ -0,0 +1,1016 @@
...
+  <section>
+    <h3 id="iframe-allow-attribute">The <code>allow</code> attribute of the
+    <code>iframe</code> element</h3>
+    <pre class="idl">
+partial interface HTMLIFrameElement {
+    [CEReactions, SameObject, PutForwards=value] readonly attribute DOMTokenList allow;
+};</pre>
+    <p><{iframe}> elements have an "<code>allow</code>" attribute, which
+    contains an unordered set of unique space-separated tokens that are ASCII
+    case-insensitive. The allowed values are names of features. Unrecognized
```
zcorpan: Why ASCII case-insensitive?

clelland: This keeps it in line with other iframe attributes, like sandbox. I didn't think it was useful to insist that these tokens are case-sensitive, while sandbox attributes on the same element are insensitive.

index.bs
clelland: Good catch; I'll fix that.

index.bs
clelland: Same here.

index.bs
clelland: The rule, which we used for origin trials as well, is so that scripts running synchronously in the document will always have a consistent, effectively-immutable policy. Without that, a script in the head could run before the meta element is encountered.

'After' is supposed to refer to tree order, but ideally I'd want elements injected via script to be ignored as well, for the same reason. I'm just not sure how to do that in the spec yet.

zcorpan: Maybe this should be an attribute on `html` instead, then? Like `manifest`.

index.bs
zcorpan: Related discussion: https://github.com/whatwg/html/issues/1665

index.bs
clelland: That's an interesting possibility -- it would only allow a single policy, and would mean that we abandon the http-equiv idea, but it would certainly be parseable before any scripts get a chance to run.

I'll remove the <meta> bits from the spec, and bring this up in an issue separately then. Having the ability to set the policy for the top-level frame purely in HTML is nice-to-have, but not, I think, strictly necessary, with the set of features currently listed.

igrigorik:
Awesome, thanks for the update! There's more stuff to flesh out here, but we can tackle those bits in separate PRs. Merging.

Issue #60, opened 2017-03-15
Feature vs W3C Process term


nigelmegitt:
The [W3C Process](https://www.w3.org/2017/Process-20170301/) uses the word "feature" in a weakly defined way that generally seems to mean "something substantive" in a Recommendation track document - using the name "Feature" here is a clash that needs to be avoided, since the two do not appear to be coincident, or they might be sometimes, but are not necessarily. 

If there is a clash between the two usages it will probably be a lower cost path to change this document than the W3C Process. On the other hand what term to use instead? "Capability" perhaps?

If there is not intended to be a clash then it would be good to clarify that the intention is that the features in the feature policy are exactly those in Recommendations and to delegate definition of those features to the Recommendations where they are defined.

A nice follow-up would be to add or reference the good practice that features should be explicitly named and defined. An example where this is done through a list of feature designators is [TTML1](https://www.w3.org/TR/ttml1/) and [TTML2](https://www.w3.org/TR/ttml2/) which include a language that allows documents to declare which features a processor needs to support in order to process that document, and for processors to declare which features they support.

zcorpan:
https://www.w3.org/2017/Process-20170301/#substantive-change defines "substantive changes" as "Corrections that do not add new features" or "New features". This seems to contradict your claim that it uses "feature" to mean "something substantive".

The describing text for "New features" is

> Changes that add a new functionality, element, etc.

The readme for this repo says

> Examples of features that could be controlled by feature policy include:
>
>* getUserMedia (Camera, Speakers and Microphone)
>* Fullscreen
>* Geolocation
>* MIDI
>* Payments
>* Synchronous XHR
>* Synchronous scripts
>* Vibrate
>* ...

I fail to see how the two are different.

Even if they are indeed subtly different, I do not understand why this would be a problem, and more generally I object to the idea that W3C specs should avoid general terms that happen to be defined in the W3C Process. For example, the term "member" has very different meaning in the W3C Process compared to "member" in WebIDL, and this has not been a problem. W3C Process also defines the term "proxy"...

nigelmegitt:
Hi @zcorpan thanks for the reply. Your references to the Process appear to back up my argument, in my view, not to contradict it! I'm confused about why our views would differ here.

When I raised the issue I did include the possibility that in fact the usages are intended to be the same, and made a proposal for clarification if that is the case.

[...]

PR #63, opened 2017-03-30
Rename potentially ambiguous term "Feature"

clelland:
This now uses the term "policy-controlled feature", but mentions that
the term "Feature" will be used as shorthand.

Fixes: #60

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -125,7 +125,7 @@ spec:fetch; type:dfn; text:value
   <h2 id="framwork">Framework</h2>
   <section>
     <h3 id="features">Policy-controlled Features</h3>
-    <p>A <dfn data-lt="feature|policy-controlled feature">policy-controlled feature</dfn>
+    <p>A <dfn export data-lt="policy-controlled feature">policy-controlled feature</dfn>
     is an API or behaviour which can be enabled or disabled in a document or web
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -124,20 +124,26 @@ spec:fetch; type:dfn; text:value
...
-    <h3 id="features">Features</h3>
-    <p>A <dfn>feature</dfn> controlled by this framework is an API or behaviour
-    which can be enabled or disabled in a document or web worker by referring to
-    it in a <a>feature policy</a>.
-    <p><a>Features</a> have a <dfn>feature name</dfn> keyword, which is a token
-    used in <a>policy directives</a>, and a <a>default allowlist</a>, which
-    defines whether the <a>feature</a> is available to top-level documents, and
-    how access to that feature is inherited by cross-origin frames.</p>
+    <h3 id="features">Policy-controlled Features</h3>
+    <p>A <dfn data-lt="feature|policy-controlled feature">policy-controlled feature</dfn>
```
zcorpan: Add `export` attribute. But I think "feature" shouldn't be exported. You can use `<a lt="policy-controlled feature">feature</a>` when linking instead.

clelland: Thanks, that makes sense. That removes "feature" from the "Terms defined by this specification" section as well.

Issue #69, opened 2017-04-28
[TAG] Interaction with New Features


slightlyoff:
[From TAG review](https://github.com/w3ctag/design-reviews/issues/159)

The default behaviour of new features added in the future must depend on the shipping status of the feature in the platform. For example, if a document.write policy is added it must "fail open" in order to avoid breaking existing websites, and therefore the only safe scenario in which a feature could be introduced with a closed default status if is if support for the feature in feature policy is a required part of shipping the feature itself. Is there a mechanism for tying together the introduction of a feature in the platform and the addition of a new token to this specification? (see also registry point)

clelland:
> if a document.write policy is added it must "fail open" in order to avoid breaking existing websites

That's correct -- if we take an existing, ubiquitous web feature, such as document.write, and put it under policy control, we need to make its default allowlist `["*"]`, or else much existing content will be broken.

> therefore the only safe scenario in which a feature could be introduced with a closed default status if is if support for the feature in feature policy is a required part of shipping the feature itself

Also true -- a default allowlist of `[]` would mean that you *have* to use feature policy in order to enable the feature in *any* context.

The middle-ground of `["self"]` can be used for new features in a couple of ways -- you can fully integrate with feature policy, or you can accept that it is available at the top-level and not in cross-origin frames (any browser implementer could do that even without implementing any support for FP), or you could fake it without supporting FP by recognizing `<iframe allow="newfeature">` to enable the feature in cross-origin frames, much like `allowfullscreen` is handled today.

> Is there a mechanism for tying together the introduction of a feature in the platform and the addition of a new token to this specification?

No, there's currently no such mechanism. We could provide guidance on this, if that is appropriate.

Adding a token to this spec should be done by reference, I think -- I'd like the feature list appendix here to be non-normative, and simply a guide to features which are defined in other specs, which any browser may or may not support. Ideally, other specs can simply say something like

    The *foo* API is a policy-controlled-feature [See FEATURE-POLICY],
    with a feature name keyword of "foo". The default allowlist is ["self"].

and then go on to define the behaviour when enabled / disabled by policy. (Exact wording TBD)

raymeskhoury:
We chatted about this in the last meeting we add. We're not quite sure the exact concern being raised. Maybe you could clarify?

Is the problem that we want to encourage people adding new features to be aware of Feature Policy - specifically:
a) To be able to add their feature to feature policy if it makes sense
[...]

PR #80, opened 2017-06-09
Add guide for integrating new features with FP

clelland:
This addresses some of the concerns brought up in #69, and provides
additional guidance to spec authors looking to integrate with feature
policy.

### DIFF
--- a/integration.md
+++ b/integration.md
@@ -1,0 +1,130 @@
+#Defining Policy-Controlled Features
+
+This guide is for authors of other standards who want to use Feature Policy as
+a mechanism for enabling and disabling their features. It's a collection of
+principles and guidelines (more than actual rules) for coming up with sensible
+defaults, whether the feature in question is a long standing feature on the
+web, or something brand new.
+
+## Referencing the Feature Policy spec
+
+There is no central registry for policy-controlled features. Feature authors who
+ wish to allow Feature Policy to control their features on the web can do so by
+referencing the Feature Policy specification.
+
+The following example shows how a feature could be declared as a
+policy-controlled feature:
+
+> Example:
+
+>###Section N: Feature Policy Integration
+> The Sample API is a *policy-controlled feature*, as defined by
+[Feature Policy](https://wicg.github.io/feature-policy/).
+
+> * The **feature name** for the Sample API is "`sample`".
+> * The **default allowlist** for the Sample API is `["self"]`.
+
+> When disabled in a document, the `getArbitrarySamples()` method MUST return a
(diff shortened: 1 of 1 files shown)


clelland:
@raymeskhoury -- does this make sense as the start of a guide for spec authors? Anything else you'd like to see mentioned?

integration.md
```diff
@@ -0,0 +1,130 @@
...
+>     <html>
+>       <head>
+>         <script>document.write('<script src="important-security-feature.js"></scr'+'ipt>')</script>
+>         ...
+>
+> then an attacker could embed the site in a frame with the `document.write`
+> feature disabled, and bypass the critical security script.
+>
+> To avoid this, if disallowed in a feature-policy-aware document, calling
+> `document.write()` will fire a `SecurityError` exception on the `document`
```
raymeskhoury: Interesting. I would be worried that people would start using feature policy without realizing this implication and open themselves up to security problems.

integration.md
```diff
@@ -0,0 +1,130 @@
...
+> object. If disallowed in a feature-policy-oblivious document, calling
+> `document.write()` feature must cause the current page to be navigated to an
+> error page which provides the option to open the page that violated the policy
```
raymeskhoury: This is good guidance. I think we will need to spec this though, at the point when legacy features actually start using this.

integration.md
```diff
@@ -0,0 +1,130 @@
...
+> productivity measure,) and wanted to make this a policy-controlled feature,
+> then the developers would have a choice -- whether to define the new feature
+> as "block cat pictures", or as "load cat pictures".
+>
+> If the goal is to block cat pictures, such that they cannot be loaded into a
+> document at all, then defining the feature as "load cat pictures" makes
+> sense, and the feature would be given a default allowlist of `[*]`. This way,
+> cat pictures would be allowed by default on the entire web, but once disabled
+> in a document, would not be available anywhere in that document, even in
+> child frames.
```
raymeskhoury: I think there are other reasons to express the feature in terms of the ability to do something:
-Double negatives are confusing and error-prone
-Consistency is a good thing to try to encourage across features.

raymeskhoury:
Overall looks really good to me :+1:

integration.md
clelland: This is a hypothetical situation; one which we can avoid by more carefully defining the behaviour of the feature when disallowed. I've rewritten some of this to try to make that clearer.

integration.md
clelland: Added, thanks!

PR #82, opened 2017-08-18
Update section on 'Other and related mechanisms'

clelland:
This now includes a discussion of the relationship between feature
policy and CSP, and between FP and iframe sandboxing.


ecki:
Maybe also mention Default Powerful Feature Permission https://www.w3.org/TR/permissions/

PR #91, opened 2017-10-19
Ask for tests for normative changes in CONTRIBUTING.md

foolip:
See https://github.com/foolip/testing-in-standards/blob/master/policy.md
for context. This phrasing matches that used for many other specs'
CONTRIBUTING.md files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,3 +21,12 @@ If you added a contributor by mistake, you can remove them in a comment with:
 
 If you are making a pull request on behalf of someone else but you had no part in designing the 
 feature, you can remove yourself with the above syntax.
+
+# Tests
+
+For normative changes, a corresponding
+[web-platform-tests](https://github.com/w3c/web-platform-tests) PR is highly appreciated. Typically,
+both PRs will be merged at the same time. Note that a test change that contradicts the spec should
+not be merged before the corresponding spec change. If testing is not practical, please explain why
+and if appropriate [file an issue](https://github.com/w3c/web-platform-tests/issues/new) to follow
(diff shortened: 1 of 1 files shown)


foolip:
@clelland, from I've seen of FP testing I think things are already in a state where this would make sense, but please correct me if I'm wrong. (Or if you don't want to, this is just a suggestion.)

clelland:
I like it, thanks!

I'll have to create the labels to match that text; are those labels generally used consistently across spec repos? (I checked a couple at random, and don't see them, but it might be just that I don't understand GitHub issue labels well enough)

foolip:
Ah, so those labels are actually in web-platform-tests:
https://github.com/w3c/web-platform-tests/labels/type:untestable
https://github.com/w3c/web-platform-tests/labels/type:missing-coverage

So the idea is that the issues would be filed in wpt. Some specs link to "ongoing work" in their headers so that this isn't entirely invisible. Example: https://dom.spec.whatwg.org/

clelland:
That makes sense :) And now I see that the link you provide is to file a WPT issue.

Should the text mention explicitly that it's talking about filing an issue on a different repo than this one?

foolip:
You mean in the case that one doesn't follow the link but just reads the text? That's pretty like, I agree. I'll suggest a change.

foolip:
(Tomorrow, gotta run. But no need to wait for me if you have something in mind.)

clelland:
I was just thinking something like "... if appropriate, [file a web-platform-tests issue] to follow up later."

foolip:
There, updated! I have a bunch of these PRs out, but I'll use this wording going forward.

clelland:
Awesome, thanks! Merged

PR #128, opened 2018-01-18
Remove exitPictureInPicture in picture-in-picture FP

beaufortfrancois:
FYI @mounirlamouri


clelland:
Thanks for checking, @mounirlamouri 

Aside from the hard-to-parse commit line :) this looks fine; merging.

Issue #125, opened 2018-01-11
Replace "feature name" with "identifier" or similar


clelland:
From https://github.com/whatwg/xhr/pull/177#issuecomment-352839570

> Introducing that as a "feature named X" and then saying X has a feature name is somewhat confusing by the way. Maybe it makes more sense to call one the feature name and the other the identifier.

It *is* confusing to say that a feature has both a name and a "feature name", where the first is the term used to refer to the feature in specs, and the second is a character string used when declaring policies.

The proposal is to replace "feature name" with something like "identifier" to avoid this confusion. I'm not sure if "identifier" is best, or if there's something more descriptive that we should use.

annevk:
The suggestion in https://github.com/whatwg/xhr/pull/177#issuecomment-356975278 to just have one of these seems pretty good. Unless there's a compelling reason to have both.

clelland:
In that case, the identifier is pretty important, so we'd be dropping the requirement that features have any other kind of name.

Spec integration text would look something like

> This specification defines a policy-controlled feature identified by the string "`foo`". Its default [allowlist](https://github.com/WICG/feature-policy/issues/103) is `'self'`.

and would then refer to the feature like

> If the responsible document is not allowed to use the "`foo`" feature, then throw a `NoFoosAllowed` `DOMException` and abort these steps.

Would that makes sense?

annevk:
Yeah, I think that would work. Assuming some of the terms there are linked to a definition.

PR #131, opened 2018-01-23
Remove feature name

clelland:
Removes the extra 'feature name' concept from the spec, and just declares the existing "feature-name" string to be an identifying token for a feature. This should simplify spec integration and remove the need for authors to define two different names for a policy-controlled feature.


clelland:
This addresses #125

PR #140, opened 2018-02-23
Remove out-of-date and incorrect ref

clelland:
The original reference to the JSON RFC was updated to point to
RFC 8269, rather than RFC 8259. However, the more correct thing
to do would have been to remove the reference entirely, because
the spec no longer uses JSON.


annevk:
Note that for JSON on the web platform I think we always want to reference JavaScript. Perhaps indirectly via https://infra.spec.whatwg.org/#json.

clelland:
Thanks, that's good to know actually.

(In this case, we're just not referencing JSON anymore at all; the link pre-dated the header syntax change, and should have been removed months ago)

PR #155, opened 2018-04-25
Explainer for 'vertical-scroll' policy.

ehsan-karamad:
This document is a brief explainer for the feature and the proposed implementation steps.

### DIFF
--- a/policies/vertical_scroll.md
+++ b/policies/vertical_scroll.md
@@ -1,0 +1,42 @@
+# Vertical Scroll Policy
+Vertical scroll policy is a feature introduced to assist websites in blocking
+certain embedded contents from interfering with vertical scrolling. Stopping a user from vertically
+scrolling the page might be a frustrating experience. This issue has been filed and tracked as a
+[bug](https://crbug.com/611982) for Chromium for a while.
+
+## Proposed Solution
+The proposed solution is introducing a new feature, namely, `vertical-scroll` which gives the priviledge of potentially
+interfering with user's scroll to a frame (the whole page or a subset of `<iframe>`s). This means if a certain `<iframe>`
+does not have the feature enabled, it should not be able to interfere nor affect vertical scrolling in its embedder.
+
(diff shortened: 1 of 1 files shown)


clelland:

yoavweiss:
This contribution seems substantive and would be best if @ehsan-karamad joins the WICG.

cc @cwilso

clelland:
I marked this (and the other commits to this folder) as non-substantive, because they aren't related to the actual feature policy spec at all; they're ideas of potential policies that are being considered.

(The guidance suggested that "_if the contribution does not concern a normative part of a specification... the contribution can be marked as non-substantive_")

OTOH, there's no harm in getting @ehsan-karamad and @loonybear to join WICG, as they are both involved in feature policy in other ways as well.

ehsan-karamad:
I submitted a request to join the group.

yoavweiss:
> (The guidance suggested that "if the contribution does not concern a normative part of a specification... the contribution can be marked as non-substantive")

Hmm, I did not understand "substantive" to necessarily mean "normative". But you are correct that this is what the guidance says in [some cases](https://labs.w3.org/hatchery/repo-manager/pr/id/WICG/IntersectionObserver/240) (but not [others](https://labs.w3.org/hatchery/repo-manager/pr/id/WICG/feature-policy/160)).

@plehegar and/or @dontcallmedom - any guidance on the correct use of "non substantive" PRs? Would be great to have a link that defines exactly what defines "substantive".

plehegar:
Royalty-Free licensing obligations are limited to Essential Claims, per https://www.w3.org/community/about/agreements/summary/ .

So, if it's not normative, it can't be essential and thus it's non-substantive for IPR.

However, while a "substantive for IPR" change is necessarily normative, not all normative changes are necessarily essential. For example, changes that reasonable implementers would not interpret as changing architectural or interoperability requirements or their implementation would not be considered substantive for IPR.

Issue #153, opened 2018-03-15
Typo on "allowusermedia" section


compulim:
In [`allowusermedia`](https://wicg.github.io/feature-policy/#iframe-allowusermedia-attribute), there is a copy-and-paste typo:

> If the iframe element has an "`allow`" attribute whose value contains the token "`payment`", then the "`allowusermedia`" attribute must have no effect.

Instead of `payment`, it should be "`camera` or `microphone`".

clelland:
Thanks -- it actually needs to be a bit more verbose than the other sections, since the intention is that "`allowusermedia`" is equivalent to `allow="camera *; microphone *`", except that specifying either `camera` or `microphone` will override the policy for just that feature, leaving the other untouched.

I'll see if I can reword it to capture that.

foolip:
@clelland looks like this wasn't automatically closed by https://github.com/w3c/webappsec-feature-policy/pull/164.

PR #164, opened 2018-05-11
Clarify what allowusermedia actually does.

clelland:
This updates the html integration section on 'allowusermedia', and
describes more clearly how it affects the generated policy for both
camera and microphone features.

Fixes: #153

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -402,14 +402,18 @@ partial interface HTMLIFrameElement {
       <h4 id="iframe-allowusermedia-attribute">allowusermedia</h4>
       <p>The "<code>allowusermedia</code>" iframe attribute controls
       access to <a method>getUserMedia()</a>.</p>
-      <p>If the iframe element has an "<code>allow</code>" attribute whose
-      value contains the token "<code>payment</code>", then the
-      "<code>allowusermedia</code>" attribute must have no effect.</p>
-      <p>Otherwise, the presence of an "<code>allowusermedia</code>" attribute
-      on an iframe will result in adding an <a>allowlist</a> of <code>*</code>
-      for each of the "<code>camera</code>" and "<code>microphone</code>"
-      features to the frame's <a>container policy</a>, when it is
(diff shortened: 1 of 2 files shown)


annevk:
These are statements of facts. Where are the corresponding implementation requirements?

clelland:
That's a good point; this text does just describe what happens when the attribute is used. The actual processing requirements for the attribute are in [8.5. Process feature policy attributes](https://wicg.github.io/feature-policy/#process-feature-policy-attributes), but aren't affected by this clarification.

I'm wondering if all of that kind of text (non-normative, since it's just describing the effects of the algorithms that are listed later) should be removed from the spec, or relegated into notes. Would that make the spec more useful from an implementer's perspective, do you think?

annevk:
I think you want to mark it clearly as being non-normative to avoid giving the impression this is what you need to turn into an implementation. I usually start out with just requirements and notes for clarifications. And then maybe at some point an introduction when things have settled down a bit.

jan-ivar:
Does `allowusermedia` add any practical value over `allow="camera"`?

If not, [no browser](https://wpt.fyi/results/html/dom/interfaces.https.html%3Finclude%3DHTML.*?label=stable&aligned=true) implements `allowusermedia` AFAIK, so maybe we can drop it?

clelland:
I think we still need it, at least as long as `allowusermedia` is in HTML.

I've looked for an issue / PR on the WHATWG repo for removing that, if it's really completely unsupported, but haven't found one yet. If there really isn't one already, I'll open it.

jan-ivar:
@clelland I'll open one.

clelland:
And to answer your earlier question, @jan-ivar, no, I don't think that `allowusermedia` adds significant value, other than for being shorthand for `allow="camera *; microphone *"`.

clelland:
Closing, as `allowusermedia` has been removed from HTML in https://github.com/whatwg/html/commit/95e72ce392c8d8afab1b85468766f41751691f64

PR #165, opened 2018-05-14
Break the circular definitions

clelland:
This fixes the issue where both this spec and the suggested changes to
the HTML spec declared that a document "has a" feature policy, but
neither spec explicitly declared what a feature policy "is" in the
absence of a document.

It also tries to clean up some of the other sloppy usage of the term "top-level
document" and "child document" to refer to documents in browsing
contexts instead, and removes the unused spec piece about documents
being "policy-aware" or "policy-oblivious".

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -139,8 +139,7 @@ spec:fetch; type:dfn; text:value
     avoid any ambiguity.</div>
     <div class="issue">This spec currently only deals with features defined in
     Documents. We should figure out how to word this to include the possibility
-    of features in Workers as well, and for feature policies to be attached to
-    Workers.</div>
+    of features and feature policies in Workers and Worklets as well.</div>
     <p><a>Policy-controlled features</a> are identified by tokens, which are
     character strings used in <a>policy directives</a>.
     <p>Each <a>policy-controlled feature</a> has a <a>default allowlist</a>,
--- a/index.bs
+++ b/index.bs
@@ -130,8 +130,8 @@ spec:fetch; type:dfn; text:value
     <h3 id="features">Policy-controlled Features</h3>
     <p>A <dfn export
     data-lt="policy-controlled feature">policy-controlled feature</dfn> is an
-    API or behaviour which can be enabled or disabled in a document by referring
-    to it in a <a>feature policy</a>.
+    API or behaviour which can be enabled or disabled in a browsing context by
+    referring to it in a <a>feature policy</a>.
     <div class="note">For brevity, policy-controlled features will often be
     referred to in this document simply as "Features". Unless otherwise
     indicated, the term "feature" refers to <a>policy-controlled features</a>.
@@ -140,8 +140,9 @@ spec:fetch; type:dfn; text:value
     <p><a>Policy-controlled features</a> are identified by tokens, which are
     character strings used in <a>policy directives</a>.
     <p>Each <a>policy-controlled feature</a> has a <a>default allowlist</a>,
-    which defines whether that feature is available to top-level documents, and
-    how access to that feature is inherited by cross-origin frames.</p>
+    which defines whether that feature is available in documents in top-level
+    browsing contexts, and how access to that feature is inherited in nested
+    browsing contexts.</p>
     <p>A user agent has a set of <dfn>supported features</dfn>, which is the set
     of <a data-lt="policy-controlled feature">features</a> which it allows to be
     controlled through policies. User agents are not required to support every
@@ -156,7 +157,7 @@ spec:fetch; type:dfn; text:value
   </section>
   <section>
     <h3 id="policies">Policies</h3>
-    <p>A {{Document}} has a <dfn>feature policy</dfn>, which consists of:</p>
+    <p>A <dfn>feature policy</dfn> is a struct with the following items:</p>
     <ul>
       <li>A set of <a data-lt="inherited policy set">inherited policies</a>.
       </li>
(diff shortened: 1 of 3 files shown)


index.bs
```diff
@@ -137,11 +137,16 @@ spec:fetch; type:dfn; text:value
     indicated, the term "feature" refers to <a>policy-controlled features</a>.
     Other specifications, defining such features, should use the longer term to
     avoid any ambiguity.</div>
+    <div class="issue">This spec currently only deals with features defined in
+    Documents. We should figure out how to word this to include the possibility
+    of features in Workers as well, and for feature policies to be attached to
+    Workers.</div>
```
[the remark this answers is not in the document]
clelland: Added, thanks.

index.bs
```diff
@@ -169,6 +174,14 @@ spec:fetch; type:dfn; text:value
   </section>
   <section>
     <h3 id="inherited-policies">Inherited policies</h3>
+    <p>An <dfn data-lt="inherited policy|inherited policies">inherited
+    policy</dfn> is either 'Enabled' or 'Disabled'.</p>
+    <p>An <dfn>inherited policy set</dfn> is an ordered map from
```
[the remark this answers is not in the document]
clelland: It really is, isn't it? :)

I've updated it to just call this the "inherited policy", akin to the "declared policy", and then declared that "inherited policy for feature X" is just getting the value for X out of the map.

index.bs
```diff
@@ -156,53 +160,53 @@ spec:fetch; type:dfn; text:value
...
+    <p>A <dfn>feature policy</dfn> is a struct with the following items:</p>
     <ul>
-      <li>A set of <a data-lt="inherited policy set">inherited policies</a>.
+      <li>An <a data-lt="inherited policy">inherited policy</a>.
       </li>
       <li>A <a data-lt="declared policy">declared policy</a>.
       </li>
     </ul>
-    <p>An <dfn>empty feature policy</dfn> has an <a>inherited policy set</a>
+    <p>An <dfn>empty feature policy</dfn> has an <a>inherited policy</a>
```
annevk: Should this be "an empty feature policy is a feature policy that has"?

index.bs
```diff
@@ -156,53 +160,53 @@ spec:fetch; type:dfn; text:value
...
   <section>
     <h3 id="inherited-policies">Inherited policies</h3>
+    <p>An <dfn data-lt="inherited policy|inherited policies">inherited
+    policy</dfn> is an ordered map from <a
+    data-lt="policy-controlled feature">features</a> to either 'Enabled' or
+    'Disabled'.</p>
+    <p>The <dfn>inherited policy for a feature</dfn> <var>feature</var> is the
+    value in the <a>inherited policy</a> whose key is <var>feature</var>. After
+    a <a>feature policy</a> has been initialized, its <a>inherited policy</a>
+    will contain an value for each <a>supported feature</a>.</p>
```
annevk: a value

index.bs
```diff
@@ -156,53 +160,53 @@ spec:fetch; type:dfn; text:value
...
+    <div class="note">
     <p>Each document in a frame tree inherits a set of policies from its parent
     frame, or in the case of the top-level document, from the defined defaults
-    for each <a>policy-controlled feature</a>. This inherited policy set
-    determines the initial state (enabled or disabled) of each feature, and
-    whether it can be controlled by a <a>declared policy</a> in the document.
+    for each <a>policy-controlled feature</a>. This inherited policy determines
+    the initial state (enabled or disabled) of each feature, and whether it can
```
annevk: 'Enabled' or 'Disabled' (seems better to spell them consistently (or change the others))

index.bs
```diff
@@ -156,53 +160,53 @@ spec:fetch; type:dfn; text:value
...
     </p>
     <p>In a {{Document}} in a [=top-level browsing context=], the inherited
-    feature set is based on defined defaults for each feature.</p>
-    <p>In a {{Document}} in a [=nested browsing context=], the inherited feature
-    set is based on the parent document's feature policy, as well as any
+    policy is based on defined defaults for each feature.</p>
+    <p>In a {{Document}} in a [=nested browsing context=], the inherited policy
+    is based on the parent document's feature policy, as well as any
     <a href="#iframe-allow-attribute">allow attributes</a> defined on the
     browsing context container.</p>
```
annevk: This isn't entirely correct. This only applies when the container is an `iframe`, right?

index.bs
```diff
@@ -854,8 +860,8 @@ partial interface HTMLIFrameElement {
               <li>Otherwise return Disabled.</li>
             </ol>
           </li>
-          <li>Otherwise, if feature is enabled in <var>parent</var> for
-	  <var>origin</var>, return Enabled.
+          <li>Otherwise, if <var>feature</var> is enabled in <var>parent</var>
+	  for <var>origin</var>, return Enabled.
```
annevk: 'Enabled'?

index.bs
clelland: I think that the way the spec is written, this should apply to all nested browsing contexts -- it would be useful for `<frames>` to have the same restrictions (as well as `<object>` when used to contain HTML docs; no idea about SVG docs) as `<iframes>`. If an API has been determined to not be usable in cross-origin `<iframes>`, you shouldn't be able to use it in a cross origin `<frame>`.

The only difference with `iframe`s, I think, is that only `iframe`s have the extra `allow`, `allowfullscreen`, `allowusermedia`, and `allowpaymentrequest` attributes. So the "Process feature policy attributes" algorithm treats iframes specially.

The Chrome implementation uses the parent's policy to define an inherited policy for *all* nested browsing contexts. but only defines `allow` on `<iframe>`.

index.bs
clelland: Changed to be consistently capitalized.

index.bs
clelland: Thanks, fixed.

index.bs
clelland: Yep. Done.

index.bs
clelland: I've created another patch, to reword this, and to change the term 'frame' where appropriate to 'nested browsing context', since 'frame' has a different meaning in HTML. (I suppose it has three or four different meanings in HTML, but they're all different than this)

https://github.com/clelland/feature-policy/pull/1

annevk: Thanks, I did mean the bit about the allow attribute. I agree that restrictions should apply regardless.

clelland:
Thanks, @annevk for the review -- merged into the spec now.

PR #170, opened 2018-05-18
Support propagating feature policy in popups.

clelland:
This change accounts for the fact that not all top-level browsing
contexts should be treated as top-level documents by feature policy. In
particular, new windows created with `window.open()` and `<a href="_blank">`
have auxiliary browsing contexts, which should get an inherited policy
based on the document that created them.


clelland:
I'm not certain what happens (given this PR) when the auxiliary browsing context is navigated. Sandboxed documents retain their sandbox flags, and that seems like the right behaviour here, although the document may be navigated to an origin where the feature would have been allowed.

clelland:
Question: Should the `sandbox propagates to auxiliary browsing contexts` flag apply to feature policy as well?

clelland:
Messed up the merge, replaced this with #259

PR #172, opened 2018-06-04
Introduce Feature policy : `document-write`

ehsan-karamad:
This document servers as an explainer of an experimental feature which was most recently committed to chromium [codebase](https://chromium.googlesource.com/chromium/src/+/d8e49fe9a8374c236010d75874c082e9701543fb). The state of the proposal is not complete as there is still a critical open questions regarding the correct implementation and behavior for cross and same-origin contents.

All suggestions and ideas are most welcomed!

### DIFF
--- a/policies/document_stream_insertion.md
+++ b/policies/document_stream_insertion.md
@@ -1,0 +1,54 @@
+# Document Stream Insertion Policy (Dynamic Mark-up)
+
+
+## Objective
+The API around document stream modification ([dynamic mark-up insertion](https://www.w3.org/TR/2011/WD-html5-author-20110705/apis-in-html-documents.html#dynamic-markup-insertion)), i.e., `document.write`, `document.writeln`, `document.open` and `document.close` 
+are anti-pattern, parser-blocking javascritp API. The objective of the proposed feature is to limit the usage of this API in websites.
+
+## Solution
+
+When the feature is disabled for a specific origin in a frame, any occurance of such API will be ignored and might
+lead to a DOMException<sup>[1](#notes)</sup>.
+
+## Usage
+The feature can be set through the HTTP headers in the response. For instance
+
+```
+Feature policy: document-stream-insertion 'none'
+```
+
+which would disable the feature for all same and cross origin content. The feature can also be set for a given `<iframe>`
+by modifying the `allow` attribute. For instance,
+
+```
+<iframe src="https://example.com" allow="document-stream-insertion http://www.google.com"></iframe>
+```
+would limit the usage of stream insertion API to only "google.com" origins using HTTPS protocol.
+
+## Open Problems
+
+The main open problem with this feature is cross-site isolation breakage. The feature provides an embedder website with the
+power to arbitrarily turn off usage of some javascript APIs in an embedded site. If the embedded site relies on loading
+important security related contents through the disabled API<sup>[2](#notes)</sup>, then the security of the embedded site
+might become compromised. Therefore, to avoid exposing sites which
+
+  * allow being embedded in other websites,
+  * rely on `document.write` for adding important scripts,
+
+to such attacks, the behavior of the feature for same-origin vs cross-origin might be different. The safest approach would be
+to either allow usage of such API when the content is cross-origin, or, disallow it by unloading the frame. This would remove
+the security concerns related to cross-site isolation.
+
+A more forgiving approach would only unload the frame if
+  * The frame is cross-origin,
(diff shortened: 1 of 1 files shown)


annevk:
I think it would be much clearer if we called this "document-write".

clelland:
Agreed -- the initial spec had a feature called `'docwrite'`, for even more terseness. Do you think `document-write` captures the idea that this also restricts `document.open` and `document.close` as well?

policies/document_stream_insertion.md
```diff
@@ -0,0 +1,54 @@
+# Document Stream Insertion Policy (Dynamic Mark-up)
+
+
+## Objective
+The API around document stream modification ([dynamic mark-up insertion](https://www.w3.org/TR/2011/WD-html5-author-20110705/apis-in-html-documents.html#dynamic-markup-insertion)), i.e., `document.write`, `document.writeln`, `document.open` and `document.close` 
+are anti-pattern, parser-blocking javascritp API. The objective of the proposed feature is to limit the usage of this API in websites.
```
clelland: sp: JavaScript

policies/document_stream_insertion.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Solution
+
+When the feature is disabled for a specific origin in a frame, any occurance of such API will be ignored and might
```
clelland: sp: occurrence (or better yet, "usage")

policies/document_stream_insertion.md
```diff
@@ -0,0 +1,54 @@
...
+Feature policy: document-stream-insertion 'none'
+```
+
+which would disable the feature for all same and cross origin content. The feature can also be set for a given `<iframe>`
+by modifying the `allow` attribute. For instance,
+
+```
+<iframe src="https://example.com" allow="document-stream-insertion http://www.google.com"></iframe>
+```
+would limit the usage of stream insertion API to only "google.com" origins using HTTPS protocol.
```
clelland: Not quite accurate: This limits it to just www.google.com, not any *.google.com origin. (and you've used HTTP, rather than HTTPS)

policies/document_stream_insertion.md
```diff
@@ -0,0 +1,54 @@
...
+  
+In other words, if a site is familiar with the feature then it is reasonable to assume that they have considered the
+consequences of `document-stream-insertion`. 
+
+#### Notes
+
+1: The _most_ correct behavior is not known yet. For now, Chrome's implementation behind the flag simply throws exception
+on API usage (for both same and cross origin). But this is not the final implementation.
+
+2: An embedded website might rely on important `<script>` added to the document using `document.write`.
```
clelland: We should note that this is already not guaranteed to work in Chrome, because of https://github.com/WICG/interventions/issues/17 -- so hopefully sites are already discouraged from implementing critical functionality via `document.write` of script content.

The difference with this proposal is that the embedder now has some control of whether or not the `write` will succeed, rather than just the user agent.

annevk:
I think it's close enough (don't forget `document.writeln()`!). I suspect the variant of `document.open()` that's actually `window.open()` won't be blocked by this? This'll require careful integration with HTML.

policies/document_stream_insertion.md
ehsan-karamad: Done, thanks!

policies/document_stream_insertion.md
ehsan-karamad: Thanks, I believe I confused site instance with origin.

policies/document_stream_insertion.md
ehsan-karamad: Thanks. Added a few lines here.

ojanvafai:
You can already do the opening behavior with window.open. Is there a specific reason to continue to allow it with document.open? I think it's pretty subtle and confusing to have the API partially function and would be better for us to keep it simple that all uses of document.open fail.

annevk:
@ojanvafai blocking both requires two checks (one for each algorithm), as they're different methods do to IDL overloading. Therefore it might also be a tad confusing, but it seems fine to block both to me.

ehsan-karamad:
Both variants of `document.open()` (returning `window` v/s `document`) are explained in the "dynamic markup insertion" section of the sepc, therefore, so far as the promise of the feature goes, we could block both. That said our current implementation in chrome (behind flag) does not block the variant returning `window`.

clelland:
Merged into master as ca24bf4

PR #174, opened 2018-06-07
Add Feature Policy for workers

odejesush:
This change adds steps for initializing workers with a Feature Policy
inherited from the owner documents.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -137,9 +137,6 @@ spec:fetch; type:dfn; text:value
     indicated, the term "feature" refers to <a>policy-controlled features</a>.
     Other specifications, defining such features, should use the longer term to
     avoid any ambiguity.</div>
-    <div class="issue">This spec currently only deals with features defined in
-    Documents. We should figure out how to word this to include the possibility
-    of features and feature policies in Workers and Worklets as well.</div>
     <p><a>Policy-controlled features</a> are identified by tokens, which are
     character strings used in <a>policy directives</a>.
     <p>Each <a>policy-controlled feature</a> has a <a>default allowlist</a>,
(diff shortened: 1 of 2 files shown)


yoavweiss:
@odejesush Could you [join the WICG](https://www.w3.org/community/wicg/participants) to appease the IPR bots? Thanks!

odejesush:
Thank you. I thought that the issue was that my GitHub account was not linked to my W3C account.

yoavweiss:
@odejesush potentially both, but joining is the first step

odejesush:
All right, I think that my account should be good to go. I'm part of the WICG and my GitHub account is linked. Thanks @yoavweiss

annevk:
I don't think we should be inheriting policies. They should declare their own. See also https://github.com/w3c/webappsec-csp/issues/336.

cc @wanderview

wanderview:
Thanks @annevk.  I think what I wrote in https://github.com/w3c/webappsec-csp/issues/336#issuecomment-422803783 would apply here as well.  Just my opinion, of course, though.

I would really like us to define some over-arching design guidelines for how inheritance should work in the web platform.  For example, origin/policies are inherited via local URLs like blob:, etc.  Clients (iframes/workers/etc) with a unique URL get their own origin/policies irrespective from their parent.  Having every API do their own thing makes for a very confusing API and brittle implementation.

I think at TPAC it was suggested we could request some guidance from the w3c TAG on this architectural issue.

clelland:
I would certainly be more comfortable with shared workers, especially, having to declare their own policies, rather than being inheriting from the page which happens to create them first.

I used to believe that dedicated workers needed to be considered subresources, but the TPAC discussions managed to convince me that it's better to be consistent and treat all workers as separate resources.

(The jury's still out on worklets, though :) )

odejesush:
Thank you for the heads up on the discussion around this @annevk and @wanderview. I subscribed to the linked issue to follow it further.

PR #180, opened 2018-06-18
Link {{Document}} in a few places

foolip:


clelland:
Merged into master as e9f14bb
Thanks!

PR #181, opened 2018-06-18
Export `Feature-Policy` as an HTTP header

foolip:
The style of &#96;Header-Name&#96; follows Fetch, but the `&#96;` in the source
are necessary because of "Markup Shorthands: css no, markdown yes" in
the &lt;pre class="metadata">.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -56,7 +56,7 @@ spec:fetch; type:dfn; text:value
     within their application. It can do so by delivering the following HTTP
     response header to define a feature policy:</p>
     <pre>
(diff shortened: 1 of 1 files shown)


clelland:
Are the literal backticks necessary? (Is there a style guide?) I would have expected that double-quotes would read more naturally.

clelland:
nm; I found https://infra.spec.whatwg.org/#byte-sequences.

Thanks for fixing that, @foolip. I'll merge, with an HTML update as well

clelland:
I updated the repo today to automatically run bikeshed on the spec and commit the rendered doc into the gh-pages branch. Merged this change into master instead of gh-pages.

PR #184, opened 2018-06-28
Add JavaScript introspection API

clelland:
This adds a section on JavaScript policy introspection, using a `policy` object defined on `document` and `iframe`.


clelland:
@domenic -- would you mind taking a look? I expect that there are many many terms which are missing xref links, but I'm more worried that I've forgotten some substantial piece.

The rendered spec is accessible at https://drafts.featurepolicy.rocks/add-introspection/ (Specifically [Section 7](https://drafts.featurepolicy.rocks/add-introspection/#introspection))

domenic:
I did a pass through, and overall this seems like the right approach! I took notes as I went, reproduced below, but they are all minor things---no missing substantial pieces I can see. I especially appreciate the intro sections and how they talked me through the observable policy concept in enough detail so I knew what was coming in the normative text.

Notes I took:

- Examples use `policy()`, but it's an attribute. Should just be `policy`.

- Attribute instead of method seems like the right choice, assuming that you return the same `Policy` object each time. Is that guaranteed? (Or, mostly guaranteed?) If it's guaranteed, then use `[SameObject]` in the Web IDL. (This is basically just documentation.)

- "the property object" is not known terminology. We have a [long-standing bug](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301) to formalize "this" in web specs. In the meantime I lean toward "this `Policy` object". We also sometimes use "the context object", but I think that's on its way out.

- Probably you should define the two getters separately, so that you can click on each IDL attribute and go to the corresponding `<dfn>`.

- The getters just say "a policy object whose associated node is that element". Probably you want to instead explicitly state something like `Each {{Document}} has a <dfn for="Document">policy object</dfn>, which is a {{Policy}} instance whose [=associated node=] is that {{Document}}`. And similarly for iframe elements. Then you can have the getters do `Return this {{Document}} object's [=Document/policy object=]`.

- It seems like you could dedupe a bit of the origin stuff into "Let _origin_ be the [appropriate origin] for this Policy object".

- Naming: all the methods are suffixed with "feature". Is that necessary? Do you anticipate non-feature-related things on this object in the future?

clelland:
> - Examples use `policy()`, but it's an attribute. Should just be `policy`.

 Thanks, fixed.

> - Attribute instead of method seems like the right choice, assuming that you return the same `Policy` object each time. Is that guaranteed? (Or, mostly guaranteed?) If it's guaranteed, then use `[SameObject]` in the Web IDL. (This is basically just documentation.)

I don't know if it matters, as long as the result of querying is the same -- but I added [SameObject] in any case, if it makes it clearer.

> - "the property object" is not known terminology. We have a [long-standing bug](https://www.w3.org/Bugs/Public/show_bug.cgi?id=27301) to formalize "this" in web specs. In the meantime I lean toward "this `Policy` object". We also sometimes use "the context object", but I think that's on its way out.

"context object" is the term I was actually thinking of; updated with your suggestion anyway, if that term is deprecated.

> - Probably you should define the two getters separately, so that you can click on each IDL attribute and go to the corresponding `<dfn>`.
>
> - The getters just say "a policy object whose associated node is that element". Probably you want to instead explicitly state something like `Each {{Document}} has a <dfn for="Document">policy object</dfn>, which is a {{Policy}} instance whose [=associated node=] is that {{Document}}`. And similarly for iframe elements. Then you can have the getters do `Return this {{Document}} object's [=Document/policy object=]`.

I updated it to state that explicitly; is there any need to even define the getters anymore?

> - It seems like you could dedupe a bit of the origin stuff into "Let _origin_ be the [appropriate origin] for this Policy object".

I've kept *declared origin* since it's also useful in *Process feature policy attributes*, and added the idea of a *default origin* for a policy. It's less duplication in the algorithms, but probably just as much text overall.

> - Naming: all the methods are suffixed with "feature". Is that necessary? Do you anticipate non-feature-related things on this object in the future?

We originally considered that this might end up as a home for introspection of other policies as well (CSP, etc) at some point, so... maybe?

We had considered something like `document.featurePolicy.allows("feature")`, but this seems more extensible.

domenic:
> I updated it to state that explicitly; 

It's never actually stated what a document/iframe's "policy object" is :). You need to explicitly state it's an instance of `Policy`.

> is there any need to even define the getters anymore?

Yeah, for sure; right now there's no information on how to implement them. You need to draw an explicit connection between the getter and the "policy object" internal slot.

> It's less duplication in the algorithms, but probably just as much text overall.

I still like it this way; I get less distracted from the main flow of each method and can just worry about the exact definition of default origin elsewhere.

> We originally considered that this might end up as a home for introspection of other policies as well (CSP, etc) at some point, so... maybe?

That's a pretty solid justification. Might be worth adding a non-normative note in the spec in case anyone else wonders.

bzbarsky:
Why are we adding new `[NoInterfaceObject]` things?

New properties on Document should probably be `[Unscopable]` to prevent compat problems.

The `policy` names are pretty generic.  Should they be `featurePolicy` instead?

clelland:
@bzbarsky, the name was deliberately generic -- eventually it could be a namespace for introspection of other policies: CSP for instance, or Referrer Policy -- rather than adding a new member to document for each.
(@domenic pointed this out from the other end in https://github.com/WICG/feature-policy/pull/184#issuecomment-401154845, that the method names have "feature" in them -- we could certainly consider moving the word "feature" from the methods to the attribute)

clelland:
Merging, though the whole section is marked as unstable. Chrome is going to experiment with this as an origin trial in versions 69-71, and incorporate feedback into this section as we learn from developers and other parties what the API should look like.

PR #187, opened 2018-07-10
Update broken WebVR spec links

Malvoz:
The [current link](https://w3c.github.io/webvr/spec/1.1/)(s) are broken, the WebVR spec is now accessible at https://immersive-web.github.io/webvr/spec/1.1/.

However these links may need to change again soon as:
> Development of the WebVR API has halted in favor of being replaced the [WebXR Device API](https://immersive-web.github.io/webxr/). Several browsers will continue to support this version of the API in the meantime.

### DIFF
--- a/features.md
+++ b/features.md
@@ -168,9 +168,9 @@ If disabled in a document, then calls to the [`vibrate()`](https://w3c.github.io
 
 ### vr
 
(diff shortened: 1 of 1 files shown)


clelland:
Thanks, @Malvoz -- I don't think it's a problem that it'll change again: we can add the XR spec as well, and there's a possibility that the feature name will change as well anyway.

clelland:

PR #188, opened 2018-07-12
Add Reporting section

paulmeyer90:
This change adds a "Reporting" section to the spec, which defines the "feature-policy-violation" report type.

The explainer is also updated.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -28,10 +28,6 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
   type: dfn
     text: report type
     text: visible to reportingobservers
-    for: report
-      text: body; url: report-body
-  type: interface
-    text: ReportBody; url: reportbody
 </pre>
 <style>
   .unstable::before {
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -22,6 +23,16 @@ spec:html; type:dfn; text:paymentrequest
 spec:fetch; type:dfn; text:name
 spec:fetch; type:dfn; text:value
 </pre>
+<pre class="anchors">
+spec:reporting; urlPrefix: https://w3c.github.io/reporting/
```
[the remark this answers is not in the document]
paulmeyer90: Hmm, all I know is that the linking fails without it. I wonder if something isn't being exported properly though.

index.bs
```diff
@@ -909,7 +972,49 @@ partial interface HTMLIFrameElement {
...
+    containing |message|, about the <a>violation</a> of the policy on
+    |feature|.</p>
+
+    1.  Let |body| be a new {{FeaturePolicyViolationReportBody}}, initialized
+        as follows:
+
+        :   [=FeaturePolicyViolationReportBody/featureId=]
+        ::  |feature|'s string representation.
+        :   [=FeaturePolicyViolationReportBody/message=]
+        ::  |message|
```
clelland: Message is not defined on FeaturePolicyViolationReportBody. I can't remember where we came down on whether a message was required or not, but the two sections here should match.

index.bs
```diff
@@ -909,7 +972,49 @@ partial interface HTMLIFrameElement {
       <li>Return "<code>Disabled</code>".</li>
     </ol>
   </section>
+  <section>
+    <h3 id="report-violation">Generate report for violation of |feature| policy
+    with |message| and |disposition|, on |settings|</h3>
+
+    <p> Given a feature (|feature|), a string (|message|), another string
+    (|disposition|), an <a>environment settings object</a> (|settings|), and an
```
clelland: I think that |disposition| should be left out of the parameter list here. Even if/when we do have a report-only mode defined, the spec calling this won't know if it's active or not. This algorithm should eventually be responsible for determining the state of that flag and setting the report body appropriately.

index.bs
```diff
@@ -909,7 +972,49 @@ partial interface HTMLIFrameElement {
...
+        :   [=FeaturePolicyViolationReportBody/message=]
+        ::  |message|
+        :   [=FeaturePolicyViolationReportBody/sourceFile=]
+        ::  null
+        :   [=FeaturePolicyViolationReportBody/lineNumber=]
+        ::  null
+        :   [=FeaturePolicyViolationReportBody/columnNumber=]
+        ::  null
+        :   [=FeaturePolicyViolationReportBody/disposition=]
+        ::  |disposition|
```
clelland: For now, let's just make this default to `"enforce"`

index.bs
```diff
@@ -909,7 +972,49 @@ partial interface HTMLIFrameElement {
...
+        [=FeaturePolicyViolationReportBody/lineNumber=], and
+        [=FeaturePolicyViolationReportBody/columnNumber=] accordingly.
+
+    3.  If |group| is omitted, set |group| to "default".
+
+    4.  Execute [[reporting/#queue-report]] with |body|,
+        "feature-policy-violation", |group|, and |settings|.
+
+    Note: This algorithm should be called when a feature policy has
+    been <dfn data-lt="violation|violated">violated</dfn>. It is up to each
```
clelland: I'd like to put the definition of "violation" in the "Reporting" section, rather than in a note here, if it's possible. We can leave the note, and have it link to that definition.

index.bs
```diff
@@ -537,6 +544,62 @@ partial interface HTMLIFrameElement {
...
+    - <dfn for="FeaturePolicyViolationReportBody">columnNumber</dfn>: If known,
+      the column number in [=FeaturePolicyViolationReportBody/sourceFile=] where
+      the <a>violation</a> occured, or null otherwise.
+
+    - <dfn for="FeaturePolicyViolationReportBody">disposition</dfn>: A string
+      indicating whether the <a>violated</a> feature policy was enforced in this
+      case. [=FeaturePolicyViolationReportBody/disposition=] will be set to
+      "enforce" if the policy was enforced, or "report" if the <a>violation</a>
+      resulted only in this report being generated (with no further action taken
+      by the user agent in response to the violation).
```
clelland: Can we add a note here that says that there is currently no mechanism for enabling a report-only mode, so all reports generated will have this field set to "enforce" ? Otherwise, people may go looking for that, to try to figure out how to enable it.

PR #201, opened 2018-08-14
Reporting: fix typos

ebidel:
Fixing some name corrections.


clelland:

ebidel:
Should the `Feature-Policy` header take a `report-to` directive? Or will violations always be reported to the default group?

PR #205, opened 2018-08-31
[lazyload.md] Update to fix typos

Malvoz:


clelland:
Thanks for catching those -- merged.

Issue #204, opened 2018-08-20
[Feature policy: animations] What does "blocking" an animation mean?


birtles:
* Do blocked animations still dispatch events?
* Does calling `Element.animate()` with a disallowed property throw?
* If an animation specifies both allowed and disallowed properties, do the allowed ones still run?

ehsan-karamad:
Thanks for the issue. A blocked animation is expected to still run but at a very reduced sampling/interpolation rate. From the explainer 

> ... These properties are "animated" by simply jumping from the initial to the final value at the midpoint of the animation.

so potentially the behavior could be an abrupt transition to the final state (perhaps in the middle or at the end of the interval).

> Do blocked animations still dispatch events?

Yes, they should; and the behind-the-flag implementation on Chrome seems to support that as expected. In tested this on a demo page [here](https://feature-policy-demos.appspot.com/animations.html?on) by listening to the `animationiteration` event.

> If an animation specifies both allowed and disallowed properties, do the allowed ones still run?

Good point...maybe just disallow the whole animation (in favor of a more consistent and potentially simpler implementation).

birtles:
> ... These properties are "animated" by simply jumping from the initial to the final value at the midpoint of the animation.

That sentence is in the context of describing how properties with an animation type of "discrete" currently animate. It doesn't define how "blocked" animations are expected to work with this policy.

Issue #203, opened 2018-08-20
[Feature policy: animations] Description of animations needs review


birtles:
The [feature policy description](https://github.com/WICG/feature-policy/blob/master/policies/animations.md) appears to have a lot of inaccuracies. e.g.

> In order to produce animations on the web, developers declare transitions inside of CSS.

Or CSS animations, or Web Animations, or SVG SMIL, or JavaScript.

> One way to avoid this scenario is to move the animations off of the CPU entirely. Modern desktop computers and mobile devices have powerful GPUs that can perform certain kinds of animations much more efficiently, and can leave the CPU free to handle other things, like responding to user input.

The animations still run on the CPU. The GPU is just used for much of the heavy-lifting of producing each animation frame.

> Many (most?) CSS properties are interpolable

Most CSS properties are not "interpolable". See how many properties have a `'discrete'` animation type in [property-list.js](https://github.com/web-platform-tests/wpt/blob/ee6c9d315b05a2afe40a13cbf26d3126d52c34a0/web-animations/animation-model/animation-types/property-list.js).

Furthermore, "interpolable" is not dependent just on the property, but on the specific values being used (e.g. `auto` cannot currently be smoothly interpolated with `100px`).

> Other properties are not interpolable -- there is no such thing as an intermediate state -- font-face, for instance, or background-image

`background-image` can be interpolated in Webkit-based engines (including Blink) using `cross-fade()`.

ehsan-karamad:
Thanks for filing these. I will get back to this thread and fix the explainer after concluding the main promise of the feature mostly discussed in issues w3c/document-policy#10 and w3c/webappsec-permissions-policy#204.

ehsan-karamad:
Reviving this thread as the animation policy has been reworked. It no longer deals with *composited* animations and is now about disabling certain *layout* animations instead. Please, take another look.

ehsan-karamad:
(apologies for the delayed response)
Thanks for the comment's and corrections. I created the pull request w3c/webappsec-permissions-policy#239 to address the mentioned issues.

AFAIK on Chrome the aforementioned animations do not always lead to re-running layout; the policy is based on assuming the worst, i.e., if the animation *can in general* cause layout changes then it will be disabled. Understandably this would block certain animations which would not have caused re-layout; however the policy is meant to be a promise on decreasing the number of layout runs on a page and ideally lead to more responsive pages.

Issue #202, opened 2018-08-20
[Feature policy: animations] Compositor accelerated animations should not be allowed


birtles:
As per my message to [blink-dev](https://groups.google.com/a/chromium.org/d/msg/blink-dev/nRNPNwlRS6E/nKWyk2XOAgAJ) I have strong concerns with this feature allowing compositor animations but disallowing others including:

* Doing so would require either adopting a lowest common denominator approach (whereby blink would be required to disallow `filter` animations, contrary to what the explainer currently states) or would produce different results in different engines (creating a Web compatibility nightmare since content that, for example, waits for a `transitionend` from a certain animation might never get it).

* Using a whitelist of properties puts the burden on site maintainers to update (or forever hold Web content back) and will quickly become unwieldy. In 2019 authors might need to write:

```html
<iframe allow="animations(transform,translate,scale,rotate,opacity,background-color,filter,offset,offset-distance,clip-path)"></iframe>
```

* The conditions that govern whether a particular animation can be run on the compositor are very subtle and change over time. As a result even if a property is part of the set of "lower common denominator" properties or an author-specific whitelist, it might not be able to be animated on the compositor. That seems to undermine the whole purpose of this feature policy which is to "ensure smooth animations".

I suggest this feature policy should restrict all declarative animations.

ehsan-karamad:
Thanks for posting here as well. I think the explainer needs revisions (as also requested in issue w3c/webappsec-permissions-policy#203).

I will try to respond to the concerns in line. Hopefully more people familiar with `animations` and web platform will chime in.
  
> Doing so would require either adopting a lowest common denominator approach (whereby Blink would be required to disallow filter animations, contrary to what the explainer currently states) 

Thanks. Based on the idea that animations policy should eventually be parametric, I believe the lowest common denominator approach is not necessarily an issue since it will be up to the developer to select the white list.  If however we go for a non-parametric version *and* common denominator, then `filter` should not be part of the policy at this time.
 
> or would produce different results in different engines (creating a Web compatibility nightmare since content that, for example, waits for a transitionend from a certain animation might never get it).

I don't think allowing for arbitrarily different behaviors across the browsers is the goal here. If an animation is disallowed it should still fire the same events but only jump from initial state to final state
 (as discussed in issue w3c/webappsec-permissions-policy#204 ). If it is allowed, it should run as expected regardless of it being actually composited.

On that note, if a website relies on something such as the *computed style during* the animation then there could be a problem.

[...]

PR #212, opened 2018-09-05
Update animations.md

ehsan-karamad:
Revised the animations policy to propose a modified policy that blocks layout inducing animations as opposed to the non-composited animations.

The changed is motivated discussions in issues #202, #203, and #204.

### DIFF
--- a/policies/animations.md
+++ b/policies/animations.md
@@ -1,86 +1,92 @@
-Allowed Animations Policy
+Allow Layout Animations Policy
 ===========
 
-The 'animations' policy-controlled feature can be used in a document or frame to
-restrict the set of CSS properties which can be animated. This can ensure smooth
-animations, by only allowing those properties which can be animated on the GPU,
-using the hardware compositor.
+The 'layout-animations' policy-controlled feature can be used in a document or frame to
+restrict the set of CSS properties which can be animated. This feature is proposed to 
+ensure certain types of animations which *may* lead to layout and potentially heavy CPU
+load are blocked.
 
 What does that mean?
 ------------
 
-In order to produce animations on the web, developers declare transitions inside
-of CSS. These transitions can be used to animate just about any CSS property --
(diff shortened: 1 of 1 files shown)


policies/animations.md
```diff
@@ -1,86 +1,92 @@
...
-
-One way to avoid this scenario is to move the animations off of the CPU
-entirely. Modern desktop computers and mobile devices have powerful GPUs that
-can perform certain kinds of animations much more efficiently, and can leave the
-CPU free to handle other things, like responding to user input.
-
-Deeper explanation
+In order to produce animations on the web, developers declare transitions of style
+in either CSS (e.g., `@keyframes` and `animation`) or JavaScript (e.g., `element.animate()`). 
+In principal, these transitions can be used to animate just about any CSS property (
```
clelland: Minor nit -- can you move the "(" to the next line to avoid unnecessary whitespace after it?

policies/animations.md
```diff
@@ -1,86 +1,92 @@
...
+backgrounds, sizes, margins, fonts, borders, positions amongst many other).
+
+In order to support such animations, the browser has to constantly update the values for
+all of these properties, and then calculate all their effects on the page; ideally, this has
+to has to be done sixty times per second.
+
+When properties such as `size` or `position` are changed in an animation, many other 
+elements on the page to have to be moved around constantly (*layout and re-layout*), and 
+browsers can struggle to keep up. When that happens, the result is a potentially stuttering
+animation and a poorly performing and non-responsive web page; sometimes even the entire page
```
clelland: >the entire page slows down

I realize that I missed a word in the original text -- I think I intended to say that in the worst case, the entire *machine* can slow down, as all CPU resources are devoted to the animation.

ehsan-karamad: page changed to *machine*

policies/animations.md
ehsan-karamad: Done.

PR #219, opened 2018-09-17
Merge parsers

clelland:
The rendered version can be seen at https://drafts.featurepolicy.rocks/merge-parsers/#parse-policy-directive


clelland:
@domenic, does this approach to optional parameters make sense here?

annevk:
You probably want to use "is given" as terminology to use on whether an optional parameter is provided.

Aside: is merging with the CSP parser still planned?

Issue #129, opened 2018-01-19
Feature Proposal: Client Hints delegation


igrigorik:
[Client Hints](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints) defines a mechanism for UA+server to negotiate a set of HTTP request header fields that can be used for proactive content negotiation. For example, the server can request that the client send a [`DPR` request header field](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-3.1) such that it can pick the right image / CSS / HTML markup.

The site author must explicitly opt-in and specify which hints they want to receive via the [Accept-CH](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-2.2.1) header — e.g. `Accept-CH: DPR` response header indicates that the UA should send the `DPR` hint. Optionally, the site can also request that the UA remember this preference for a specified period of time, via [Accept-CH-Lifetime](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-2.2.2) header.

---

One existing limitation that came up in [privacy review](https://github.com/httpwg/http-extensions/issues/372#issuecomment-350803441) is that CH does not expose 'delegation controls' for 1P vs 3P resources; CH does not provide a means for 1P to specify which 3P's should be allowed to request various hints. I believe (and propose :)) that Feature Policy is the right tool for the job to address this: we can define each hint as a [policy-controlled feature](https://wicg.github.io/feature-policy/#policy-controlled-feature). 

For example, enumerating existing hints today...

- Hint: [DPR](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-3.1) → FP: `ch-dpr` with default policy of `self`.
- Hint: [Width](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-3.2) → FP: `ch-width` with default policy of `self`.
- Hint: [Viewport-Width](https://tools.ietf.org/html/draft-ietf-httpbis-client-hints-04#section-3.3) → FP: `ch-viewport-width` with default policy of `self`.
- Hint: [Save-Data](https://wicg.github.io/netinfo/#save-data-client-hint-request-header-field) → FP: `ch-save-data` with default policy of `*`.

[...]

PR #220, opened 2018-09-24
“Should request be allowed to use feature” algorithm

eeeps:
This PR is part of an attempt to address #129 – using Feature Policy to let authors opt-into sending specific Client Hints to specific origins.

The overall, basic idea is:

1. A new set of policy-controlled features – one for each Client Hint header field – is defined in the Client Hints spec.
2. A new algorithm which determines whether a request is able to use a policy-controlled feature is defined in the Feature Policy spec. That's defined in this PR.
3. The step in Fetch that appends Client Hint headers calls the new algorithm, checking if a request is allowed to use a Client Hints’ policy-controlled feature before appending the Client Hint header.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1188,6 +1188,24 @@ partial interface HTMLIFrameElement {
     Note: This algorithm should be called when a feature policy has
     been <a>violated</a>.
   </section>
+  <section>
+    <h3 id="should-request-be-allowed-to-use-feature">Should <var>request</var> be allowed to use <var>feature</var>?</h3>
+    <p>Given a feature (<var>feature</var>) and a  <a for="/">request</a> (<var>request</var>), this algorithm returns <code>true</code> if the request should be allowed to use <var>feature</var>, and <code>false</code> otherwise.</p>
+    <ol>
+      <li>Set |global object| to |request|’s <a for="request">client</a>’s <a for="environment settings object">global object</a>.</li>
+      <li>Set |document| to null.</li>
+      <li>If |global object| is a {{Document}}, set |document| to |global object|.</li>
+      <li>If |global object| is a {{Window}}, set |document| to |global object|’s <a>associated `Document`</a>.</li>
+      <li>If |document| is null, return <code>false</code>.
+        <div class="issue">How can we enable secure use of policy-controlled features within requests initiated by clients that do not have documents? e.g., from within {{WorkerGlobalScope}}s or {{WorkletGlobalScope}}s, i.e., within Service Workers?</div>
+      </li>
+      <li>Let |origin| be |request|’s <a for="request">URL</a>’s <a for="url">origin</a>.</li>
+      <li>Let |result| be the result of executing <a href="is-feature-enabled">Is feature enabled in document for origin?</a> on |feature|, |document|, and |origin|.
+      </li>
+      <li>If |result| is "<code>Enabled</code>", return <code>true</code>.</li>
+      <li>Otherwise, return <code>false</code></li>
+    </ol>
+  </section>
 </section>
 
 <section>


index.bs
```diff
@@ -1188,6 +1188,24 @@ partial interface HTMLIFrameElement {
     Note: This algorithm should be called when a feature policy has
     been <a>violated</a>.
   </section>
+  <section>
+    <h3 id="should-request-be-allowed-to-use-feature">Should <var>request</var> be allowed to use <var>feature</var>?</h3>
+    <p>Given a feature (<var>feature</var>) and a  <a for="/">request</a> (<var>request</var>), this algorithm returns <code>true</code> if the request should be allowed to use <var>feature</var>, and <code>false</code> otherwise.</p>
+    <ol>
+      <li>Set |global object| to |request|’s <a for="request">client</a>’s <a for="environment settings object">global object</a>.</li>
```
igrigorik: Hmm, I'd love to get @annevk to sanity check this.. 

Given that we're talking about passed in "request", shouldn't we use use request's attribute directly, as defined by Fetch? E.g. Fetch#client or [Fetch#request-window](https://fetch.spec.whatwg.org/#concept-request-window)?

index.bs
```diff
@@ -1188,6 +1188,24 @@ partial interface HTMLIFrameElement {
...
+      <li>Set |document| to null.</li>
+      <li>If |global object| is a {{Document}}, set |document| to |global object|.</li>
+      <li>If |global object| is a {{Window}}, set |document| to |global object|’s <a>associated `Document`</a>.</li>
+      <li>If |document| is null, return <code>false</code>.
+        <div class="issue">How can we enable secure use of policy-controlled features within requests initiated by clients that do not have documents? e.g., from within {{WorkerGlobalScope}}s or {{WorkletGlobalScope}}s, i.e., within Service Workers?</div>
```
igrigorik: ServiceWorker's client has an associated origin, we should be checking against that?

@jakearchibald any guidance? :)

index.bs
jakearchibald: Depends what you want to happen in non-window contexts like workers.

jakearchibald: A request's client may be null, so this spec has to deal with that.

index.bs
```diff
@@ -1188,6 +1188,24 @@ partial interface HTMLIFrameElement {
     Note: This algorithm should be called when a feature policy has
     been <a>violated</a>.
   </section>
+  <section>
+    <h3 id="should-request-be-allowed-to-use-feature">Should <var>request</var> be allowed to use <var>feature</var>?</h3>
+    <p>Given a feature (<var>feature</var>) and a  <a for="/">request</a> (<var>request</var>), this algorithm returns <code>true</code> if the request should be allowed to use <var>feature</var>, and <code>false</code> otherwise.</p>
+    <ol>
+      <li>Set |global object| to |request|’s <a for="request">client</a>’s <a for="environment settings object">global object</a>.</li>
+      <li>Set |document| to null.</li>
+      <li>If |global object| is a {{Document}}, set |document| to |global object|.</li>
```
jakearchibald: The global is never a document. It's either a `Window` or a `WorkerGlobalScope`. https://html.spec.whatwg.org/multipage/webappapis.html#realms-settings-objects-global-objects

index.bs
jakearchibald: The CSP model would probably work well here. https://www.w3.org/TR/CSP3/#html-integration

Documents and worker globals have the CSP list. Requests are blocked before reaching the service worker if they fail the document's CSP. However, the service worker may also have its own CSP which blocks additional requests.

CSP also checks responses. I don't know if that's desirable here.

index.bs
eeeps: In what contexts? If there’s no other way to get an origin in those contexts, I guess we'll return false.

index.bs
eeeps: Oops. Thanks!

index.bs
eeeps: Workers having their own Feature Policy is being figured out here: https://github.com/WICG/feature-policy/issues/207. Until that's resolved, I think we can/should just return false in Worker contexts.

index.bs
```diff
@@ -1053,6 +1053,23 @@ partial interface HTMLIFrameElement {
...
   </section>
+  <section>
+    <h3 id="should-request-be-allowed-to-use-feature">Should <var>request</var> be allowed to use <var>feature</var>?</h3>
+    <p>Given a feature (<var>feature</var>) and a  <a for="/">request</a> (<var>request</var>), this algorithm returns <code>true</code> if the request should be allowed to use <var>feature</var>, and <code>false</code> otherwise.</p>
+    <ol>
+      <li>Set |window| to |request|’s <a for="request">window</a>.</li>
+      <li>If |window| is not a {{Window}}, return <code>false</code>.
+       <div class="issue">Feature Policy within `no-window` contexts ({{WorkerGlobalScope}} or {{WorkletGlobalScope}}) is being figured out in <a href="https://github.com/WICG/feature-policy/issues/207">issue #207</a>. After that’s resolved, update this algorithm to allow fetches initiated within these contexts to use policy-controlled features. *Until* that’s resolved, disallow all policy-controlled features (e.g., sending Client Hints to third parties) in these contexts.</div>
+       <div class="issue">Could this algorithm be run before fetching starts, when |request|’s <a for="request">window</a> is still the string `'client'`?</div>
```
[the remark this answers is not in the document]
eeeps: Done.

clelland:
This looks good, thanks for the ping @yoavweiss

PR #229, opened 2018-10-12
Fix inherited policy origin

clelland:
This corrects an issue with the "Define an inherited policy for feature"
algorithms where the wrong origin was used to compare against the
container policy and the parent document's feature policy. The origin of
the document in the frame (or of the document potentially loaded into
that frame) should be used, rather than the parent's origin.

This also cleans up some of the linked terms in the various algorithms
involved in this.


clelland:
Rendered preview at https://drafts.featurepolicy.rocks/fix-inherited-policy-origin/

PR #240, opened 2018-11-01
Updating image policies

loonybear:


policies/optimized-images.md
```diff
@@ -239,104 +138,94 @@ For an `<img>` element, if neither the width or the height of the source image e
...
-"example.com"
-<img id="normal-size" src="test.png">
-<img id="oversized" src="test-oversized.png">
+
+**Future Development**
+
+Image formats affect file size. We want to support different default values for different image formats.
+We want to allow developers to specify the parameters as well. In practice, they would look something like this:
+```html    
+<iframe allow="unoptimized-images(BMP(1, 0.5), JPG(2, 0.2))"></iframe>
```
clelland: We should probably use the MIME types of image/bmp and image/jpeg (or just 'bmp' or 'jpeg') here instead of having to define a new mapping of strings to image formats.

policies/optimized-images.md
```diff
@@ -239,104 +138,94 @@ For an `<img>` element, if neither the width or the height of the source image e
...
     ```html
-    <iframe allow="image-compression(12)"></iframe>
-    ```
-    That would apply a policy in which the maximum compression ratio allowed is set to 12.
-
-    Feature policies combine in subframes, and the minimum value of the compression ratio will be applied, so if that frame embedded another, which the syntax:
-
-    ```html
-    <iframe allow="maximum-downscaling-image(15)"></iframe>
+    <iframe allow="unoptimized-images(1.5, 0.4)"></iframe>
```
clelland: Did we decide somewhere to use KB as the unit? Would it be more natural to specify # bytes instead?
(I'm not sure, for instance, whether 1KB = 1000 bytes or 1024, or if 0.2 would mean 200 bytes, or 204, or even 205 if rounded up)

policies/optimized-images.md
```diff
@@ -239,104 +138,94 @@ For an `<img>` element, if neither the width or the height of the source image e
...
-test.png: 150px X 150px
-<img id="within-range" width="100" height="100" src="test.png">
-<img id="over-width" width="50" height="100" src="test.png">
-```
-   </td>
-  </tr>
 </table>
 
-For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width or height), the image will be rendered correctly; if the height the source image exceeds the limit, the image will be rendered with inverted colors.
+For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width or height), the image will be rendered correctly; if the height the source image exceeds the limit, the image will be rendered as placeholder image.
```
clelland: ditto here too

policies/optimized-images.md
```diff
@@ -14,222 +14,121 @@ Optimized image policies are aiming to solve problems with sites shipping images
...
-<img id="within-range" width="100" height="100" src="test.png">
-<img id="over-height" width="100" height="50" src="test.png">
-```
+<img src="resources/max-ds-img-enabled0.png" width="80%">
    </td>
   </tr>
 </table>
 
-For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width or height), the image will be rendered correctly; if the width the source image exceeds the limit, the image will be rendered with inverted colors.
+For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width or height), the image will be rendered correctly; if the width the source image exceeds the limit, the image will be rendered as placeholder image.
```
clelland: ditto here

policies/optimized-images.md
```diff
@@ -14,222 +14,121 @@ Optimized image policies are aiming to solve problems with sites shipping images
...
-<img id="within-range" width="100" height="100" src="test.png">
-<img id="over-width-and-height" width="50" height="50" src="test.png">
-```
+<img src="resources/max-ds-img-enabled1.png" width="80%">
    </td>
   </tr>
 </table>
 
-For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width of height), the image will be rendered correctly;  if both the width and the height of the source image exceed the limit, the image will be rendered with inverted colors.
+For an `<img>` element, if neither the width or the height of the source image exceeds the number of pixels allowed by the policy in the container (by default, 2 times of its container's width of height), the image will be rendered correctly;  if both the width and the height of the source image exceed the limit, the image will be rendered as placeholder image.
```
clelland: nit: "as a placeholder image"

policies/optimized-images.md
loonybear: OK. But the MIME types are sometimes img/jpg image/jpg and stuff so it is hard to include all here. I guess it'd make it challenging to implement too.

policies/optimized-images.md
loonybear: makes sense. I used 1024 but yea we should be using bytes.

PR #259, opened 2018-11-30
Support propagating feature policy in popups.

clelland:
This change accounts for the fact that not all top-level browsing
contexts should be treated as top-level documents by feature policy. In
particular, new windows created with `window.open()` and `<a href="_blank">`
have auxiliary browsing contexts, which should get an inherited policy
based on the document that created them.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -945,6 +945,24 @@ partial interface HTMLIFrameElement {
     <ol>
       <li>Let <var>context</var> be <var>document</var>'s <a>browsing
       context</a>.</li>
+      <li>If <var>context</var> is an [=auxiliary browsing context=]:
+        <ol>
+          <li>Let <var>opener</var> be <var>context</var>'s [=opener browsing
+          context=]'s [=active document=].</li>
+          <li>If either <var>context</var> is not <a
+          lt="disown its opener">disowned</a>, or <var>opener</var>'s <a>sandbox
+          propagates to auxiliary browsing contexts flag</a> is set:
+            <ol>
+              <li>Let <var>origin</var> be <var>opener</var>'s [=origin=]</li>
+              <li>If <a href="#is-feature-enabled"><var>feature</var> is enabled
+              in <var>opener</var> for <var>origin</var></a>, return
+	      "<code>Enabled</code>".
+              </li>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -945,6 +945,24 @@ partial interface HTMLIFrameElement {
     <ol>
       <li>Let <var>context</var> be <var>document</var>'s <a>browsing
       context</a>.</li>
+      <li>If <var>context</var> is an [=auxiliary browsing context=]:
+        <ol>
+          <li>Let <var>opener</var> be <var>context</var>'s [=opener browsing
+          context=]'s [=active document=].</li>
+          <li>If either <var>context</var> is not <a
```
ehsan-karamad: IIUC, this suggestions means that a *disowned* auxiliary browsing context with a sandbox-ed opener (that has the sandbox flag to propagate sandbox to popups set) will *not* inherit feature policy state. This sounds good but is a different behavior from what sandbox currently does and hence I wonder if it will block (https://github.com/WICG/feature-policy/blob/master/sandbox.md) in its current form.

Should we instead special case sandbox policies here, or for now make feature policy follow sandbox guidelines when sandbox is present.

ehsan-karamad: Taking another look I realized this instruction is suggesting what I mentioned above; to give an example,
Suppose feature `sync-xhr` is disabled inside a browsing context. Then the following table maps the combination of `sandbox` and `rel` to the feature state in the auxiliary context (i.e., results of `document.policy.allowsFeature('sync-xhr')`):

| `<iframe>` has:  \\ `window` is:  	|   `rel='noopener'`	|  `rel=''` 	|
|---	|---	|---	|
|  `sandbox` present 	|  `false` 	|   `false`	|
|  `sandbox` present, `allow-popups-to-escape-sandbox` present	|  `true` 	|  `true` 	|
|  `sandbox` not present 	|  `true` 	|  `false` 	|

clelland:
After a number of discussions on this topic, I'm closing this; the consensus right now seems to be that the features which are best suited for feature policy are those which are reset in popups anyway -- permissions, powerful features, etc, for which any requests for access can be attributed to a new top-level origin, and visible window.

PR #285, opened 2019-03-07
fix typo in header name

cvazac:
`Feature policy` => `Feature-Policy`


clelland:
Good catch, @cvazac -- merged.

Issue #260, opened 2018-12-04
export anchor terms for algorithms


frivoal:
The https://wicg.github.io/feature-policy/#is-feature-enabled algorithm (and maybe others) are useful to refer to form outside this spec. In addition to giving it a section title, giving it an exported `<dfn>` would make cross linking easier from bikeshed and the like.

clelland:
Should all of the algorithms be exported? The only ones that I think are currently used as entry points from outside specs are
 - Initialize document’s Feature Policy
 - Initialize document’s Feature Policy from response
 - Is feature enabled in document for origin?
 - Generate report for violation of feature policy on settings
 - Should request be allowed to use feature?

The others could be internal to this doc, I think.

annevk:
I'd try to be conservative with exporting and only do so when requested (or have identified a need) as there's no good tooling for updating downstream dependencies (or even identifying them).

clelland:
SGTM. I'll add *Is Feature Enabled* and *Generate Report*, as those are the two which are expected to be used externally. The *Initialize* methods are used from HTML, but it has its own biblio entries for them.

PR #287, opened 2019-03-25
Use dfn tags for algorithms, and export as appropriate.

clelland:
The four algorithms which are not referenced internally are all exported as entry points for use by other specs:

* Initialize document’s Feature Policy
* Initialize document’s Feature Policy from response
* Is feature enabled in document for origin?
* Should request be allowed to use feature?

Closes: #260


clelland:
I changed the structure of the algorithm markup to use a <dfn> for each one. I exported the algorithms which are not referenced internally (because bikeshed complains otherwise, and they *are* used as external entry points).

Issue #284, opened 2019-03-05
HTML calls Initialize document’s Feature Policy seems wrong


annevk:
When HTML creates a browsing context it calls this algorithm, but this expects the document to already participate in some hierarchy that hasn't yet been established. Either _creator_ needs to be passed in or this needs to work entirely differently.

clelland:
To be clear, is this a result of the document's browsing context being a nested browsing context with no browsing context container yet?

(I was looking at the [Implement the sandboxing](https://html.spec.whatwg.org/#implement-the-sandboxing) algorithm, which is called from the same place, and it seems to also assume the parent-child browsing context relationship, so I'm trying to disentangle the exact differences to see how to best fix this)

annevk:
Yeah; I guess that needs patching too. 😟

clelland:
Part of the issue appears to be the use in the FP spec of "nested browsing context" as though it is a subclass, or specific type of browsing context, which is incorrect. There is only one 'type' of browsing context, although a browsing context may be the nested browsing context of a specific browsing context container, which is not the case at the point in the "[Create a new browsing context](https://html.spec.whatwg.org/multipage/browsers.html#creating-a-new-browsing-context)" algorithm when "[Initialize document's Feature Policy](https://w3c.github.io/webappsec-feature-policy/#initialize-for-document)" is called.

On the HTML side, the "[Create a new browsing context](https://html.spec.whatwg.org/multipage/browsers.html#creating-a-new-browsing-context)" steps should be adjusted to allow for this; sandboxing has the same terminology issue; I've just opened whatwg/html#4444 to address these issues.

Here, the spec should be changed to use the correct terminology, and go through the browsing context container as appropriate, rather than start with the condition "if *x* is a nested browsing context, ..."

PR #291, opened 2019-03-26
Correct uses of the term 'nested browsing context'

clelland:
The term 'nested browsing context' refers to a specific attribute of a
browsing context container, and is not a kind of browsing context. This
corrects the use in the spec, and refers to 'child browsing context'
instead, or else 'nested browsing context of a browsing context
container'.

Fixes: #284


clelland:
Well, this plus https://github.com/whatwg/html/pull/4449 fix #284, I think.

PR #293, opened 2019-04-03
Update and rename lazyload.md to loading-frame-default-eager.md

ehsan-karamad:
The policy is renamed accordingly.


policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,46 @@
+loading-frame-default-eager Policy
+===========
+
+The `loading-frame-default-eager` policy-controlled feature can be used for a main frame or a nested
+`<iframe>` to overwrite the default or assigned value of the newly
+introduced `loading` [attribute](https://github.com/whatwg/html/pull/3752) for `<iframe>` and `<img>`.
```
clelland: ... and ``? Weren't we discussing a separate policy for images?

policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,46 @@
...
+this issue but changing a browser's default decision for `loading` behavior.
+
+Proposed Solution
+------------
+The `loading` attribute can take one of the three values:
+  * **`eager`** indicating that the resource should load eagerly (i.e., load even when below the fold).
+  * **`auto`** the browser makes the call for the loading behavior; it probably defaults to `eager`.
+  * **`lazy`**: the resource loads lazily.
+
+The proposed `loading-frame-default-eager` policy will affect the `auto` attribute value in that when disabled inside a document, all the `<iframe>`'s which do not explicitly set the `loading` attribute to `eager` will most probably load lazily. This is equivalent to overriding the `loading` attribute of `auto` with `lazy`.
```
clelland: I'd switch these two things around -- when disabled, auto (or unset) will be treated as lazy, which means that frames which don't set the attribute to `eager` will (usually) be lazily loaded.

policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,46 @@
...
+  
+This feature could be enforced either in the HTTP header or by using the `allow` attribute of an inline frame.
+
+Using the Feature
+-------------
+
+This feature can be introduced with the HTTP headers. For instance,
```
clelland: Can we also add an header example for `none`, as I think that's the case that will be useful most often, as opposed to setting specific origins, and allowing on `self`?

policies/loading-frame-default-eager.md
ehsan-karamad: This part defines what `loading` attribute is; which is related to both `` and `<iframe>`.

policies/loading-frame-default-eager.md
ehsan-karamad: Added an example. PTAL.

policies/loading-frame-default-eager.md
clelland: We shouldn't need to explain `loading` -- this feels like it should be the introduction to `loading-frame-default-eager`, so I parsed it incorrectly as "feature can be used ... to overwrite the default value ... for `<iframe>` and ``"

policies/loading-frame-default-eager.md
clelland: That's great, thanks!

policies/loading-frame-default-eager.md
ehsan-karamad: Newest version does not name ``.

policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,57 @@
...
+
+Similarly, the feature could be set through the `allow` attribute of an inline frame:
+```HTML
+<iframe src="https://example.com" loading="auto" allow="loading-frame-default-eager 'none'"></iframe>
+```
+which regards all the `<iframe>`s with `loading="auto"` or `<iframe>`s that do not set `loading` attribute (hence default to `auto`) as lazily loaded frames. Note that the frame itself is not affected by the policy and possibly loads eagerly.
+
+The Extra Mile
+-----------
+In general the feature could allow an expanded set of enforcement policies with the use of [parametric features](https://github.com/WICG/feature-policy/issues/163). Under a geralized `loading-frame` policy, the following potential values could be supported (in the order of permissiveness):
```
Malvoz: > geralized

generalized?

policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,57 @@
...
+This feature could be enforced either in the HTTP header or by using the `allow` attribute of an inline frame.
+
+Using the Feature
+-------------
+
+This feature can be introduced with the HTTP headers. For instance,
+```HTTP
+Feature-Policy: loading-frame-default-eager 'none'
+```
+would cause all the nested `<iframe>`'s and the corresponding nested and auxiliary browsing contexts to load lazily, unless the have specifically set `loading="eager"`. For example, if the document has
```
Malvoz: > unless the have specifically set `loading="eager"`

unless _they_  have specifically set `loading="eager"`?

policies/loading-frame-default-eager.md
ehsan-karamad: Thanks.

policies/loading-frame-default-eager.md
ehsan-karamad: Thanks!

policies/loading-frame-default-eager.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `loading-frame-default-eager` policy-controlled feature can be used for a main frame or a nested
+`<iframe>` to override the default behavior of the newly
+introduced `loading` [attribute](https://github.com/whatwg/html/pull/3752).
+
+What does that mean?
+------------
+With the `loading-frame-default-eager` attribute, developers could prioritize the loading of different inline frames on a web page. This however could become a cumbersome process and not quite scalable for larger
+web sites specially given that applying the attribute is origin-agnostic. The proposed policy aims to resolve
+this issue but changing a browser's default decision for `loading` behavior.
```
Malvoz: > The proposed policy aims to resolve
> this issue but changing a browser's default decision

"_by_ changing a browser's..."?

Malvoz: These typos also exist in https://github.com/w3c/webappsec-feature-policy/pull/295/files :(

policies/loading-frame-default-eager.md
ehsan-karamad: Done. Thanks.

policies/loading-frame-default-eager.md
ehsan-karamad: Done thanks. and yes it should definitely be sharing the same typos which I will tend to tomorrow.

PR #304, opened 2019-04-30
Create focus-without-user-activation.md

ehsan-karamad:
Add an explainer for `focus-without-user-activation` policy which helps reduce abusing automatic focus in embedded contents.

### DIFF
--- a/policies/focus-without-user-activation.md
+++ b/policies/focus-without-user-activation.md
@@ -32,7 +32,7 @@ In a nutshell:
   * Around step 4 of the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autofocus for `autofocus` the algorithm should return if the policy `focus-without-user-activation` is disabled and the algorithm is not
   [triggered by user activation](https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation).
   * Before starting [steps](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) for `element.focus(options)` the same verification for the policy and user activation should be performed.
-  * Aroudn step 2 of the the [spec](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) for `window.focus(options)` the same enforcement should be made (using the browsing context of the `window` itself to obtaint he feature policy state.
+  * Aroudn step 2 of the the [spec](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) for `window.focus()` the same enforcement should be made (using the browsing context of the `window` itself to obtaint he feature policy state.
 
 Using the Feature
 -------------
--- a/policies/focus-without-user-activation.md
+++ b/policies/focus-without-user-activation.md
@@ -1,0 +1,57 @@
+focus-without-user-activation Policy
+===========
+
+The `focus-without-user-activation` policy-controlled feature helps control the use of
+automated focus in a main frame or `<iframe>`.
+
+What does that mean?
+------------
+Automatic focus could happen through:
+  * Use of `autofocus` attribute on a form control (e.g. `<input>`),
+  * Use of scripted focus such as `element.focus()` and `window.focus()`.
+
+Automatic focus is potentially
+[problematic](https://github.com/w3c/webappsec-feature-policy/issues/273) since it provides bad
+embedded content with a tool to steal input focus from the top-level. The proposed feature provides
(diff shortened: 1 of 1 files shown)


policies/focus-without-user-activation.md
```diff
@@ -0,0 +1,57 @@
...
+  should not change.
+
+
+Using the Feature
+-------------
+This feature can be introduced with the HTTP headers. For instance,
+```HTTP
+Feature-Policy: focus-without-user-activation 'none'
+```
+would cause all all automatic focus in the page (and neste contexts) to fail unless it has been
```
marian-r: Redundant "all", "neste" -> "nested".

policies/focus-without-user-activation.md
```diff
@@ -0,0 +1,57 @@
...
+triggered by user activation.
+
+To disable the feature for a specific `<iframe>`, the `allow` attribute can be used:
+```HTTP
+Feature-Policy: focus-without-user-activation 'self'
```
marian-r: Was there an intention to provide an example with `<iframe allow="..">`? Maybe it got lost when copy/pasting.

policies/focus-without-user-activation.md
```diff
@@ -0,0 +1,57 @@
...
+initialized through user activation. This essentially means that `autofocus` will be disabled (
+unless a new element is inserted, with `autofocus`, as a result of user gesture). The scripted focus
+will also only work if it has started with user gesture.
+
+Details on "disabling focus"
+------------
+To avoid unnecessary breakage of existing contents, disabling the focus should still let parts of
+the [focus update steps](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps)
+run. In a nutshell:
+  * The `activeElement` of the blocked document should be updated. All the focus related events
```
marian-r: I'm not en expert on focus behavior in browsers, but does that mean that `activeElement` would be updated even though it actually would not be active?

ehsan-karamad: This might be a bit hand-wavy and I still need to dig through the steps in the spec, but what I am suggesting is that the document inside the `<iframe>` that is blocked shouldn't know about the focus not really working. I wonder if it makes sense for all the steps in focus work (inside the `<iframe>`) but not make it target for keyboard input (focusable area).

policies/focus-without-user-activation.md
ehsan-karamad: Thanks!

policies/focus-without-user-activation.md
ehsan-karamad: Added an example.

policies/focus-without-user-activation.md
marian-r: I'm not sure what would be the best approach. But out of curiosity: Would you want the document not to know about focus being blocked for security reasons? Anyway, PR looks good to me.

policies/focus-without-user-activation.md
```diff
@@ -25,16 +25,14 @@ will also only work if it has started with user gesture.
...
-  should not change.
-
+All automated focus eventually call into the [focus update steps](https://html.spec.whatwg.org/multipage/interaction.html#focus-update-steps) algorithm. When the policy
+is disabled, this algorithm should not run.
+
+In a nutshell:
+  * Around step 4 of the [spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-autofocus for `autofocus` the algorithm should return if the policy `focus-without-user-activation` is disabled and the algorithm is not
+  [triggered by user activation](https://html.spec.whatwg.org/multipage/interaction.html#triggered-by-user-activation).
+  * Before starting [steps](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) for `element.focus(options)` the same verification for the policy and user activation should be performed.
+  * Aroudn step 2 of the the [spec](https://html.spec.whatwg.org/multipage/interaction.html#dom-window-focus) for `window.focus()` the same enforcement should be made (using the browsing context of the `window` itself to obtaint he feature policy state.
```
marian-r: `Aroundn` -> `Around`, `obtaint he` -> `obtain the`, closing parentheses at the end.

ehsan-karamad: Thanks you! I fixed those two and also updated the focus algorithm link (update focus steps -> focusing steps).

JSteunou:
Can't wait to see this flag usable everywhere :)

Issue #148, opened 2018-03-13
commas and semicolons both allowed, with different rules for overriding


dbaron:
If I execute (in my head, so error-prone!) the algorithm for [8.2. Parse header from value and origin](https://wicg.github.io/feature-policy/#parse-header), it splits the value on commas, and for each later piece, merges it in by adding only the features not already present.  Thus earlier ones override later ones.

But the thing it does with the values is call [8.3. Parse policy directive from value and origin](https://wicg.github.io/feature-policy/#parse-policy-directive), which splits the value on semicolons, and replaces earlier entries with later ones.

I think, together, this means that in:

```
Feature-Policy: vibrate https://a.example.com/; vibrate https://b.example.com/, vibrate https://c.example.com/
```

the resulting policy is `vibrate https://b.example.com/`.

This seems pretty weird to me.  Is it intended?  If it is, perhaps there should be some prose explaining why?

annevk:
(This should also be a shared code path with CSP.)

briansmith:
As I mentioned in #314, HTTP header parsing rules mean that the following mean that:

```
Feature-Policy: x
Feature-Policy: y
```

means the same thing as:

```
Feature-Policy: x, y
```

This is why CSP uses semicolons to separate directives within a single policy.

I agree with @annevk that, in the absence of a strong motivation otherwise, it should work like CSP.

clelland:
@briansmith, by 'work like CSP', do you mean that we intersect the policies if multiple headers (or multiple comma-separated declarations) are found?

I think that would just involve changing [Merge directive with declared policy](https://w3c.github.io/webappsec-feature-policy/#merge-directive-with-declared-policy) to do something more complex when an allowlist for a given feature has already been declared.

I'd like to keep the property that:
```
Feature-Policy: feature1 allowlist1
Feature-Policy: feature2 allowlist2
```
is the same as 
```
Feature-Policy: feature1 allowlist1; feature2 allowlist2
```

but if there are better ways to handle
```
Feature-Policy: feature1 allowlist1
Feature-Policy: feature1 allowlist2
```

then we should explore those.

briansmith:
> @briansmith, by 'work like CSP', do you mean that we intersect the policies if multiple headers (or multiple comma-separated declarations) are found?

Yes.

[...]

Issue #55, opened 2017-02-23
Support for declaring feature policy in HTML


clelland:
In the most recent updates to the explainer and spec, there are two ways to set a policy on a document:

1. `Feature-Policy` HTTP header
2. `allow*` attributes on `<iframe>` elements

The HTTP header is the only way for a document to declare its own policy (the iframe attributes are, of course, set by the embedding document), and is the only way to override the policy on the top-level document.

Since not every web page author has the ability to control the headers sent with their documents, it would be good to have a way to set the declared policy in HTML.

Proposals
========

Support the `<meta http-equiv>` alternative to HTTP headers in HTML
-----------------------------------------------------

We can treat `<meta>` elements as equivalent to HTTP headers, and allow document authors to include them in `<head>` elements. This has the advantage of being consistent with the Origin Trials mechanism, as well as with other `http-equiv` meta tags. We have to be careful in this case to not allow the tag to take effect after scripts have had a chance to run, though, as the policy enforced by the document should be effectively immutable, as far as scripts are concerned. Practically, I think this means we need to do one of two things:

1. Don't allow the policy to have any effect if it appears after `<script>` or `<link>` elements, or if inserted via script. This might cause problems in browsers that can run scripts (like Chrome extensions) that aren't included in the `<head>`, since it may not be possible to include them early enough.

2. Require that the policy be delivered within the first 1024 bytes of the response, so that user agents can scan for it before starting to parse the document, like `<meta charset>`. This isn't a scalable solution if other header-ish mechanisms want to adopt it as well; at some point, it's not going to be possible to cram everything into the first kilobyte.

Make the in-document declared feature policy an attribute on the `<html>` tag
-----------------------------------

This means that there can be only one such declared policy (which may be a good thing), and forces it to be declared on the very first element in the document. If very large, it might push a `<meta charset>` declaration out of the first kilobyte, but we might be able to define a shorthand format for such declarations. (If it doesn't look like it's trying to be an HTTP header, then maybe it doesn't need to have the same format)

Other ideas?
-------
I'm certainly open to other ideas here.

zcorpan:
(`<meta charset>` isn't the only way to specify encoding: UTF-8 BOM works as well as in-file encoding declaration, and of course `charset` parameter in `Content-Type` header works.)

clelland:
That's true, but if you have full control over your headers, then it may be just as easy to include a `Feature-Policy` header as well.

annevk:
[...]

PR #313, opened 2019-05-16
HTML <meta>-delivery

eeeps:
Addresses #55 

This is a first, naive, stab at this. I've copied some language from https://www.w3.org/TR/CSP3/#meta-element and given it a once over.

At the very least, it needs a corresponding PR to HTML, which I'll link to here, as soon as it's in.


annevk:
Does this have interest from multiple implementers?

eeeps:
@annevk I think the only browser folks in #55 were Googlers (who seemed positive) and yourself (who expressed concerns). So unless I've missed conversations elsewhere: no, not yet.

index.bs
```diff
@@ -427,6 +427,55 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
...
+    <div class="example">
+      <pre highlight="html">
+        &lt;meta http-equiv="Feature-Policy" content="fullscreen 'none'; geolocation 'none'"&gt;
+      </pre>
+    </div>
+
+    Implementation details can be found in HTML's <a>Feature Policy state</a>
+    `http-equiv` processing instructions [[!HTML]].
+
+    Authors are <em>strongly encouraged</em> to place <{meta}> elements as early
```
Malvoz: Is it important that the `<meta charset>` comes before `<meta http-equiv>`s? I asked the same thing in https://github.com/w3c/webappsec-csp/issues/387.

index.bs
```diff
@@ -427,6 +427,55 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
...
+      In the above, the call to req.open will succeed, because the policy
+      disabling sync-xhr occurs after it, in the markup. 
+
+    </div>
+
+    Note: A policy specified via a <{meta}> element will be enforced along with any other policies active, regardless of where they're specified. However, once a specific Feature Policy directive has been set by way of an HTTP header or <{meta}> element, that directive cannot be further updated or changed.
+    <!--TODO this is first wins. do we want last wins? -->
+
+    Note: Modifications to the <{meta/content}> attribute of a <{meta}> element after the element has been parsed will be ignored.
+
```
briansmith: What about adding a new `<meta http-equiv='Feature-Policy'>` header to the DOM using script? This is done regularly with CSP to tighten the policy after the initial scripts have loaded.

index.bs
```diff
@@ -427,6 +427,55 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
...
+        &lt;meta http-equiv="Feature-Policy" content="sync-xhr ‘none"&gt;
+      </pre>
+
+      In the above, the call to req.open will succeed, because the policy
+      disabling sync-xhr occurs after it, in the markup. 
+
+    </div>
+
+    Note: A policy specified via a <{meta}> element will be enforced along with any other policies active, regardless of where they're specified. However, once a specific Feature Policy directive has been set by way of an HTTP header or <{meta}> element, that directive cannot be further updated or changed.
+    <!--TODO this is first wins. do we want last wins? -->
```
briansmith: How multiple `Feature-Policy` headers should be combined is the subject of #148. I don't think this PR should change anything other than adding text to say that using `<meta>` for Feature Policy is allowed, and each such `<meta>` is interpreted as though a separate `Feature-Policy` header field were appended to the HTTP response header. Then the text that resolves #148 can explain how multiple 'Feature-Policy' header fields are combined together.

annevk: I don't think that's possible as they are parsed independently. If you start a quoted string in one and end inside a `<meta>` that should not work.

clelland: Good point :)

A `<meta>` element would have to be parsed and merged separately from the response headers. 

That's also necessary because it would be encountered after [create a new browsing context](https://html.spec.whatwg.org/multipage/browsers.html#creating-a-new-browsing-context) has already initialized the document's policy (unless we specify that it interacts with the preload scanner somehow, but I think there is legitimate hesitancy to adding more dependencies on that).

index.bs
eeeps: Good question, and thanks for the heads up. I'll plan to monitor the conversation there and copy any changes over.

index.bs
eeeps: This language doesn't say anything about what happens when new elements are inserted. As I plan to re-use as much of the specific HTML processing step language from CSP as possible, I expect that the behavior will be similar. 

Because FP is more about client side behavior, and less about loading/blocking resources, I don't expect that pattern to be as common here, as it is with CSP.

That said, if we do expect it to be  a common pattern, it may merit a mention in this high-level-guideline-y, non-normative bit of text.

Thoughts?

index.bs
briansmith: > > and each such `<meta>` is interpreted as though a separate `Feature-Policy` header field were appended to the HTTP response header.
>
> I don't think that's possible as they are parsed independently. 

I think you're probably interpreting me too literally. I just meant that each `<meta http-equiv=Feature-Policy>` should be interpreted the same as if it were a HTTP header.

More generally, it seems bad that CSP and Feature-Policy as specifying so many details for `<meta http-equiv>`; it seems like HTML5 itself should provide more infrastructure for these specs to reuse.

annevk: Well, for starters there's disagreement on whether we should add more `<meta http-equiv>`. I don't think it's a good idea as it makes all these policies dynamic and there's usually subtly different processing models across browsers leading to issues.

Then, the value-space is different, headers contain bytes, attribute values contain strings. Then, headers can be parsed as a single-block (and ideally combined, then parsed), elements are typically parsed case-by-case.

There are some reusable bits there, but all of it requires quite a bit of care.

And then, for consistency across `<meta http-equiv>`, it's unclear to what extent that exists today, but if someone were to invest time in that maybe some abstraction is feasible as you suggest.

eeeps:
@annevk I think that this and https://github.com/whatwg/html/pull/4670 are now at a place where it makes sense to critique specifics. Does the flow here make sense? Are there weirdnesses resulting from the fact that these pragma directives are parsed and set later/differently, that I haven't considered?

In general, being able to set Feature Policies from HTML will allow the large group of authors who do not have the ability to modify HTTP headers to use the spec. Doing so in a pragma directive seems to enable re-using as much syntax and parsing as possible, even if there are subtle differences and potential problems.

PR #314, opened 2019-05-16
Change semicolons separating directives in examples to commas.

eeeps:
Two examples in the spec show multiple directives in the same policy header, separated by semicolons. But https://w3c.github.io/webappsec-feature-policy/#algo-parse-header splits headers on commas.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -80,7 +80,7 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
     within their application. It can do so by delivering the following HTTP
     response header to define a feature policy:</p>
     <pre>
-      <a http-header>Feature-Policy</a>: fullscreen 'none'; geolocation 'none'</pre>
+      <a http-header>Feature-Policy</a>: fullscreen 'none', geolocation 'none'</pre>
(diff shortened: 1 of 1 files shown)


briansmith:
I think there should be an issue about how to deal with multiple "Feature-Policy:" headers first.

In particular, note that this:

```
Feature-Policy: x
Feature-Policy: y
```

means the same thing as:

```
Feature-Policy: x, y
```

You are implicitly implying here that multiple `Feature-Policy` header fields should be UNIONed together, but in Content Security Policy multiple header fields are INTERSECTed together. It's not clear that UNION is the right thing.

This is why CSP uses semicolons to separate directives within a single policy.

eeeps:
Found https://github.com/w3c/webappsec-feature-policy/issues/148, which clarifies why both are allowed, the differences, and what happens if there are multiples of each.

Hm. Will have to give this a think in the morning.

clelland:
This is fairly out-of-date now; since the header was renamed to `Permissions-Policy` and switched to a structured format, the comma-vs-semicolon issue has been resolved (commas won). In other contexts, like the `allow` attribute, only semicolons were ever allowed.

PR #318, opened 2019-05-23
Clarify forgiving approach for document-write policy

foolip:


foolip:
I think the intended meaning was inverted.

clelland:
I think you're right; a word got dropped there. Thanks for catching it.

PR #319, opened 2019-05-29
Fix wrong property name for reporting

sisidovski:
I tried to see reporting via ReportingObserver, but documentation seems setting wrong property name. `featureId` is correct, isn't it?


clelland:
Thanks for catching that -- yes, `featureId` is the actual property name in the reports.

PR #321, opened 2019-06-05
Fix spec link in focus-without-user-activation Policy

foolip:
Also whitespace.


clelland:
Thanks, @foolip -- that was definitely broken :(

PR #324, opened 2019-07-12
Expose new algorithms to create a Feature Policy before document is c…

dtapuska:
…reated.

Adjust algorithms so that the HTML spec will be able to create a feature
policy before the document is created. Instead of creating the document
first and then applying it afterwards. The document based algorithms
will remain until the HTML spec is changed.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -958,7 +958,7 @@ partial interface HTMLIFrameElement {
     <a>Feature Policy</a></p>
     <ol>
       <li>Let <var>policy</var> be the result of running <a>Create a Feature Policy for a browsing
-      context from response</a> given <var>document</var>'s <a>browsing context</a>, and <var>document</var>'s
+      context from response</a> given <var>document</var>'s <a>browsing context</a>, <var>document</var>'s
       <a>origin</a>, and <var>response</var>.</li>
       <li>
         Set <var>document</var>’s <a href="https://html.spec.whatwg.org/multipage/dom.html#concept-document-feature-policy">feature policy</a> to <var>policy</var>.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -909,58 +908,75 @@ partial interface HTMLIFrameElement {
       policy <var>inherited policy</var> and declared policy <var>declared
       policy</var>.
       </li>
+      <li>Return <var>policy</var>.</li>
+    </ol>
+  </section>
+  <section>
+    <h3 id="algo-initialize-for-document"><dfn export
```
clelland: This can also eventually be removed, I hope.

index.bs
```diff
@@ -909,58 +908,75 @@ partial interface HTMLIFrameElement {
...
     </ol>
   </section>
   <section>
-    <h3 id="algo-define-inherited-policy"><dfn>Define an inherited policy for
-    <var>feature</var> in <var>document</var></dfn></h3>
-    <p>Given a feature (<var>feature</var>) and a <a>Document</a>
-    (<var>document</var>), this algorithm returns the <a>inherited policy</a>
-    for that feature.</p>
+    <h3 id="algo-define-inherited-policy"><dfn id="define-inherited-policy">Define an inherited policy for
+    <var>feature</var>, and <a>origin</a> in <a>browsing context</a></dfn></h3>
```
clelland: This should just be `<dfn...>Define an inherited policy for <var>feature</var> in <a>browsing context</a></dfn>` to match the name used in create-for-browsingcontext.

index.bs
```diff
@@ -909,58 +908,75 @@ partial interface HTMLIFrameElement {
...
+    <p>Given a feature (<var>feature</var>), an <a>origin</a> (<var>origin</var>), and
+    a <a>browsing context</a> (<var>context</var>), this algorithm returns the
```
clelland: Can this be renamed to `<var>browsingContext</var>` to match the other algorithms? (Or rename the others to *context*; either way works, but consistency looks better)

index.bs
dtapuska: Yes.. That is the desire. A 3 phase patch...
1) Make algorithms available in this spec
2) Change HTML spec to call them
3) Remove dead algorithms in spec

index.bs
```diff
@@ -887,19 +887,18 @@ partial interface HTMLIFrameElement {
...
   </section>
   <section>
-    <h3 id="algo-initialize-for-document"><dfn export
-    id="initialize-for-document">Initialize <var>document</var>'s Feature
-    Policy</var></dfn></h3>
-    <p>Given a {{Document}} object (<var>document</var>), this algorithm
-    initialises <var>document</var>'s <a>Feature Policy</a></p>
+    <h3 id="algo-create-for-browsingcontext"><dfn export id="create-for-browsingcontext">Create a
+    Feature Policy for a browsing <var>context</var></dfn></h3>
+    <p>Given a <a>browsing context</a> (<var>browsingContext</var>), and <a>origin</a>
```
clelland: Super-minor nit: "and *an* origin"

PR #325, opened 2019-07-16
Remove algorithms no longer needed

dtapuska:
The HTML spec was updated in https://github.com/whatwg/html/pull/4772

Now the deprecated algorithms can be removed.


dtapuska:
We can merge this once the HTML spec change lands.

clelland:
Agreed; holding off on merge until then. LGTM otherwise.

PR #328, opened 2019-07-24
Create document-policy-explainer.md

clelland:
Add a first draft of an explainer for document policies

### DIFF
--- a/document-policy-explainer.md
+++ b/document-policy-explainer.md
@@ -95,9 +95,9 @@ A document like this:
   <body>
     <iframe src="https://a.example.com/" policy="no-unsized-media"></iframe>
     <iframe src="https://b.example.com/"
-            policy="image-compresson;bpp=1"></iframe>
+            policy="image-compression;bpp=1"></iframe>
     <iframe src="https://c.example.com/"
-            policy="image-compresson;bpp=4"></iframe>
+            policy="image-compression;bpp=4"></iframe>
   </body>
 </html>
 ```
@@ -128,13 +128,17 @@ threshold value is still 'bpp=4'.)
 
 ### Sandboxing nested content (Traditional sandbox)
 
+As examples of different ways to sandbox content using a combination of the
+`sandbox` and `policy` attributes, the following could all be used to create a
+sandboxed iframe:
+
 ```html
 <iframe sandbox>
 <iframe sandbox="allow-scripts allow-forms">
-<iframe policy="no-scripts">
-<iframe sandbox policy="scripts"
-<iframe sandbox="no-scripts" policy="scripts">
-<iframe policy="no-same-origin">
+<iframe policy="no-scripts"> (This is a non-sandboxed frame with no scripting allowed)
+<iframe sandbox policy="scripts"> (Should this be allowed?)
+<iframe sandbox="allow-scripts" policy="no-scripts"> (This one needs some work)
+<iframe policy="no-same-origin"> (This is a frame with an opaque origin, but no other sandboxing)
 ```
 
 None of these result in a `Sec-Required-Feature-Policy` header, as all of the
@@ -228,7 +232,7 @@ Some policies are boolean switches, and can be turned off with the syntax
 Others take parameters, whose types and names are specific to each policy:
 
 * `image-compression;min-bpp=2`
-* `frame-loading:lazy`
+* `frame-loading;lazy`
 
 All of the directives can be combined with commas into a single header. In
 Structured-header-speak, this is a parameterized list.
--- a/document-policy-explainer.md
+++ b/document-policy-explainer.md
@@ -1,0 +1,337 @@
+# Feature Policy: Document Policies
+
+This is a proposal for an extension to Feature Policy to cover those kinds of
+features which don't involve delegation of permission to trusted origins;
+features which are more about configuring a document, or removing features
+(sandboxing) from a document or a frame.
+
+## Start with Examples!
+
+### Performance guardrails
+
+The simplest example is a site which wants to enforce some performance
+best-practices on their own content. They can do this by serving their HTML
+content with this HTTP header:
+
+```http
+Document-Policy: no-unsized-media, no-document-write,
+                 image-compression;bpp=2, frame-loading;lazy
+```
+
+A document served with this header may embed other content, first- or
+third-party, and that content will not be subject to those restrictions. That
+content may include its own `Document-Policy` header, but the headers do not
+combine in any way.
+
+### Enforcing performance guardrails on embedded content
+
+In this example, the top level document wants to ensure that the content loaded
+into a particular frame uses best practices regarding its images. All images
+should have declared sizes, and should be reasonably compressed. It includes an
+iframe tag like this:
+
+```html
+<iframe src="https://img.example.com/"
+        policy="no-unsized-media,image-compression;bpp=2">
+```
+
+When the document is loaded over the network, the request will look like this
+(abbreviated):
+
+```http
+GET / HTTP/1.1
+Host: img.example.com
+Sec-Required-Document-Policy: no-unsized-media,image-compression;bpp=2
+```
+
+The response will need to have a conforming header in order to be loaded, so a
+document served with this header would work:
+
+```http
+Document-Policy: image-compression;bpp=1.5,no-document-write,no-unsized-media
+```
+
+While either of these headers would cause a network error to be returned
+instead:
+
+```http
+Document-Policy: no-unsized-media,image-compression;bpp=3
+Document-Policy: image-compression;bpp=2
+```
+
+If the document is successfully loaded, any subsequent navigations from that
+frame will also have their requests sent with the same
+`Sec-Required-Document-Policy` header.
+
+### Deeper nesting
+
+As above, If the top-level document uses the policy attribute in an iframe
+element, like this:
+
+```html
+<iframe src="https://img.example.com/" policy="image-compression;bpp=2">
+```
+
+Then the request for the frame will look like this:
+
+```http
+GET / HTTP/1.1
+Host: img.example.com
+Sec-Required-Document-Policy: image-compression;bpp=2
+```
+
+If the framed document then includes its own subframes, they will be sent with
+the same `Sec-Required-Document-Policy` header, as the policy set by the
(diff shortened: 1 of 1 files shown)


document-policy-explainer.md
```diff
@@ -0,0 +1,337 @@
...
+the required policy indicated by those frames will combine with the required
+policy imposed on the document itself.
+
+A document like this:
+
+```html
+<html>
+  <body>
+    <iframe src="https://a.example.com/" policy="no-unsized-media"></iframe>
+    <iframe src="https://b.example.com/"
```
Malvoz: Iframe's missing closing tag here

document-policy-explainer.md
```diff
@@ -0,0 +1,337 @@
...
+top-level document subsume the requirements on the nested frame, so the combined
+threshold value is still 'bpp=4'.)
+
+### Sandboxing nested content (Traditional sandbox)
+
+```html
+<iframe sandbox>
+<iframe sandbox="allow-scripts allow-forms">
+<iframe policy="no-scripts">
+<iframe sandbox policy="scripts"
```
Malvoz: this iframe's missing the closing tag

document-policy-explainer.md
```diff
@@ -0,0 +1,337 @@
...
+"`no-<feature-name>`":
+
+* `no-popups`
+* `no-scripts`
+* `no-document-write`
+
+Others take parameters, whose types and names are specific to each policy:
+
+* `image-compression;min-bpp=2`
+* `frame-loading:lazy`
```
eeeps: Should this be a semicolon instead of a colon?

document-policy-explainer.md
```diff
@@ -0,0 +1,337 @@
...
+A document like this:
+
+```html
+<html>
+  <body>
+    <iframe src="https://a.example.com/" policy="no-unsized-media"></iframe>
+    <iframe src="https://b.example.com/"
+            policy="image-compresson;bpp=1"></iframe>
+    <iframe src="https://c.example.com/"
+            policy="image-compresson;bpp=4"></iframe>
```
Malvoz: "Compression" in `image-compression`  is spelled "compresson" in these 2 frame policies

document-policy-explainer.md
clelland: Yes, thanks!

document-policy-explainer.md
clelland: Fixed, and added some text here as well that I had originally intended to.

document-policy-explainer.md
clelland: Fixed, thanks

clelland:
@annevk, do you want to take a look, and see if this is a good starting point for discussion?

@ojanvafai @foolip @jpchase FYI as well

huchenlei:
In deeper nesting section's last line:
```http
GET / HTTP/1.1
Host: c.example.com
Sec-Required-Document-Policy: image-compression;bpp=2
```
(Note that in the last example, the stricter requirements imposed by the top-level document subsume the requirements on the nested frame, so the combined threshold value is still **'bpp=4'**.)

I suppose the bpp=4 should be bpp=2 to match what happens in the example.

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+document which it is served with. You can use it to set things like restrictions
+on image sizes or compression ratios, lazy frame and image loading policies, use
+of synchronous APIs, or use of APIs that fall under iframe sandboxing,
+traditionally.
+
+That header doesn't impose any restrictions on content which the document embeds
+at all, and in the base case, every document is free to set its own policy
+independently.
+
+(In a minute, we'll talk about required policies, though -- and not conforming
```
foolip:
```suggestion
(In a minute, we'll talk about required policies, though – and not conforming
```

Seems like GitHub markdown doesn't convert `--` to anything.

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+* `image-compression;min-bpp=2`
+* `frame-loading;lazy`
+
+All of the directives can be combined with commas into a single header. In
+Structured-header-speak, this is a parameterized list.
+
+### Requiring a policy in your embedded content
+
+It is possible to enforce that content you embed conforms to a specific policy
+as well. This can be done in two ways -- either through a different response
```
foolip:
```suggestion
as well. This can be done in two ways – either through a different response
```

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+acknowledgement will be advertised with a `Sec-Required-Document-Policy` request
+header.
+
+If the content which is returned doesn't have a `Document-Policy` header which
+is at least as strict as the `Sec-Required-Document-Policy`, then it will not be
+loaded into the frame, and a network error will result instead.
+
+The `Sec-Required-Document-Policy` header will be present for all requests for
+documents in a frame, at every nesting level. If the framed document adds its
+own `Require-Document-Policy header`, or uses the policy attribute on its own
```
foolip:
```suggestion
own `Require-Document-Policy` header, or uses the policy attribute on its own
```

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+iframes, then the requirements will be combined in deeper requests, to ensure
+that all content meets all of the requirements being imposed on it.
+
+### Extending the Iframe Sandboxing Mechanism
+
+All existing sandbox features can be defined as document policies as well. The
+iframe `sandbox` attribute will disable all of them by default, but they can be
+turned back on either with the `policy` attribute or the `sandbox` attribute.
```
foolip: What would it look like to turn sandbox flag back off with the `policy` attribute? Would it be `<iframe sandbox policy="forms">`, as an equivalent to `<iframe sandbox="allow-forms">`?

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+(Alternately, we can define a central safelist of features which do not require
+acknowledgement; the decision comes down to whether that call is made by the
+feature policy spec or the specs which define the individual features)
+
+As a starting point for discussion, all existing sandbox flags would not require
+acknowledgement, and all other document policies would.
+
+In documents generated from `data:` URLs, and in iframe srcdoc documents, where
+the parent document controls the content of the child explicitly, and there is
+no HTTP network transfer, the acknowledgment will simply be implied.
```
foolip: How about same-origin iframes?

document-policy-explainer.md
clelland: Done

document-policy-explainer.md
clelland: Done

document-policy-explainer.md
clelland: Done

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+
+```http
+GET / HTTP/1.1
+Host: img.example.com
+Sec-Required-Document-Policy: image-compression;bpp=2
+```
+
+If the framed document then includes its own subframes, they will be sent with
+the same `Sec-Required-Document-Policy` header, as the policy set by the
+top-level document applies to all content within its frame.
```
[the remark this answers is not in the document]
clelland: That makes more sense, thanks :)

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+Document-Policy: image-compression;bpp=2
+```
+
+If the document is successfully loaded, any subsequent navigations from that
+frame will also have their requests sent with the same
+`Sec-Required-Document-Policy` header.
+
+### Deeper nesting
+
+As above, If the top-level document uses the policy attribute in an iframe
```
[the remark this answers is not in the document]
clelland: Thanks, fixed :)

document-policy-explainer.md
clelland: I think that would be the syntax, if we can do it. I'm worried about the compat implications of allowing *both* `<iframs sandbox="allow-forms">` and `<iframe sandbox policy="forms">` though.

document-policy-explainer.md
clelland: That's a possibility -- does it open up any new kinds of attacks, where being able to affect the policy on an iframe and not requiring it to consent enables things you couldn't do otherwise?

document-policy-explainer.md
foolip: Do you mean the risk of people adopting `<iframe sandbox policy="forms">` before it works in all browsers, and ending up not having the behavior they want in some browsers? There's also the potential confusion of having two equivalent ways to express two things. What's the strongest argument for having this new form? Tradeoffs :)

document-policy-explainer.md
foolip: I'm not to sure if there are plausible attacks here, but it does seem like the same argument as for `data:` and `srcdoc` applies: the same origin controls both things.

This all reminds me of the decision to not require `<iframe allow="fullscreen">` for the same-origin case, although it's just analogous and risks might be different.

document-policy-explainer.md
```diff
@@ -0,0 +1,341 @@
...
+<iframe sandbox>
+<iframe sandbox="allow-scripts allow-forms">
+<iframe policy="no-scripts"> (This is a non-sandboxed frame with no scripting allowed)
+<iframe sandbox policy="scripts"> (Should this be allowed?)
+<iframe sandbox="allow-scripts" policy="no-scripts"> (This one needs some work)
+<iframe policy="no-same-origin"> (This is a frame with an opaque origin, but no other sandboxing)
+```
+
+None of these result in a `Sec-Required-Feature-Policy` header, as all of the
+existing sandbox features are safelisted.
```
annevk: I think we should leave this as an open issue as surfacing existing sandboxing features seems quite valuable and gives servers more control over potential attacks. (In that case there would not have to be matching response header, but that seems okay, quite similar to CORS.) Also, this should be `Sec-Required-Document-Policy`.

clelland: Thanks for catching the name :)

I was looking to maintain some kind of parity between the items which are advertised in that header and the policies which you *must* declare in order to be loaded. I wonder if we'd need some different syntax to denote policies which will be enforced, but which you don't get to refuse?

Leaving an open issue seems like the right thing here, agreed.

PR #329, opened 2019-08-06
Fix broken link to Page Lifecycle's feature-policies

Malvoz:
Replace https://wicg.github.io/page-lifecycle/spec.html#feature-policies with https://wicg.github.io/page-lifecycle/#feature-policies.


clelland:
Thanks!

PR #333, opened 2019-08-19
Explain why we need more than the `loading` attribute

Malvoz:
As described in https://github.com/w3c/webappsec-feature-policy/issues/193#issuecomment-522531706 - add a sentence identifying that it is the `loading` attribute that may be "cumbersome" to use (and thus we need these policies). Because it currently reads as if it is using `loading-{frame|image}-default-eager` that may be cumbersome.

/cc @clelland, @ehsan-karamad


clelland:
Thanks -- that reads better now :)

PR #334, opened 2019-08-22
Add explicit [Exposed]

saschanaz:
Required after heycam/webidl#423. (Found from web-platform-tests/wpt#18382)


jpchase:
The two blocker issues (w3c/reporting#131 and w3c/reporting#173) have been resolved, so this can be merged now?

clelland:
Yes, this can be merged, though it should also add Worker exposure as well. I'll do that as a separate commit.

PR #338, opened 2019-09-01
webauthn => publickey-credentials

Malvoz:
Change `webauthn` to `publickey-credentials` per @equalsJeffH's request in https://github.com/w3c/webappsec-feature-policy/pull/336 (that PR didn't change the name of the feature, only the markdown link name).


clelland:
Good catch; merging.

PR #344, opened 2019-10-12
Move `publickey-credentials` from proposed to standardized feature

Malvoz:
Since the feature has been "declared in a published version of the respective specification", moving this feature to the standardized features table.


clelland:
Thanks!

PR #345, opened 2019-10-21
Add `battery` feature identifier for the Battery Status API

Malvoz:
https://w3c.github.io/battery/#feature-policy-integration


clelland:
Thanks!

PR #360, opened 2020-01-10
Remove obsolete speaker policy

jpchase:
The speaker policy was listed as proposed, but decided not to be pursued in https://github.com/w3c/mediacapture-output/issues/63#issuecomment-539068488. The partial implementation in Chrome was removed in crbug.com/985892.


clelland:
Thanks, @jpchase! (and for following up on the other documentation repos as well)

Issue #244, opened 2018-11-14
Maintain a registry


annevk:
Coming from the land of WHATWG, I don't see a problem with a centralized list as we can update standards easily. Having it will help a lot in review, maybe help with consistency, etc.

Opening this as #241 got closed without such a registry.

ojanvafai:
What concretely are you picturing here? Would a markdown file in this repo mapping names to specs be sufficient?

sideshowbarker:
> Would a markdown file in this repo mapping names to specs be sufficient?

I think that should be sufficient.

More specifically: speaking as someone who’s had need to process data from previous registries we’ve created, I think a file in version-controlled in git that people would need to file PRs to update would be better than the other wiki-based mechanisms we tried in the past (none of which worked out well).

Further, given that it’s data, I think we should make it JSON rather than markdown.

ojanvafai:
What's the goal of having a registry? I'm not pushing back, just trying to understand. For example, if we're going for human readable markdown makes sense. If it needs to be computer readable, JSON could make more sense.

sideshowbarker:
> What's the goal of having a registry?

For me, one concrete goal is to add support to the HTML checker for checking the value of the `allow` attribute for the `iframe` element, and for checking the value of `Feature-Policy` header.

Specifically, the HTML checker could check that the `allow` value or `Feature-Policy` header didn’t have a misspelled feature identifier — e.g., `geo-location` or something instead of `geolocation`.

The HTML checker already does similar checking of the `content` value of the `meta http-equiv="content-security-policy"` element and the Content-Security-Policy header (though in that case it’s done without the use of any registry).

But I can imagine other purposes for a Feature-Policy registry that wouldn’t need the data as JSON. And for the purposes of the HTML checker, I could of course parse the necessary data out of a markdown file. Having it as JSON would just make it easier.

ojanvafai:
I don't have strong opinions FWIW. Either way seems reasonable to me.

sideshowbarker:
I notice now there’s already an existing markdown document at https://github.com/WICG/feature-policy/blob/master/features.md which it seems like could serve as the basis for a registry.

annevk:
> What's the goal of having a registry?

Basically having some place to lookup everything that's in use to get a sense what defining a Feature Policy requires and ensure some kind of consistency is enforced. Ideally the specification would maintain a table of sorts.

frivoal:
[...]

Issue #75, opened 2017-04-28
[TAG] New Registry; Strategy for Updates?


slightlyoff:
[From TAG review](https://github.com/w3ctag/design-reviews/issues/159)

This spec introduces a new registry of names. What is the strategy for maintaining consistency between the [Permissions API registry](https://w3c.github.io/permissions/#permission-registry) and naming and the feature names in this spec?

raymeskhoury:
There has been some agreement that Permissions in the Permissions API registry should be a subset of features in Feature Policy. I'm not sure what the best way to enforce that is though?

clelland:
I mentioned this in #69, but I'll restate here -- I'd love to avoid introducing a new registry if possible. With the exception of `"fullscreen"` and `"allowpaymentrequest"`, which require special handling for compat reasons, I've tried to avoid mentioning any specific features in the normative text of the spec.

The features are all mentioned in an appendix, which I think should eventually just be collected from references in other specs. If other specs can define *xyz* to be a policy-controlled-feature, declare its default allowlist and define its behaviour when enabled and disabled, they should be able to do that by reference to this spec, and then this spec doesn't have to be the authoritative source for all features. Is such a thing possible?

clelland:
Ping @slightlyoff, @raymeskhoury -- Do you think that what we have is sufficient, or is an actual registry unavoidable here?

raymeskhoury:
I'm not actually sure on the spec side. Conceptually it seems ok to me not to have a registry though.

clelland:
Ping @slightlyoff  -- is this issue okay to close now?

annevk:
> Is such a thing possible?

If there's no order-dependent processing of any kind that could work, but it makes reviewing rather hard as it's unclear what's being controlled.

And note if you ever add an API you'll likely have to add all of them due to enums and such.

PR #366, opened 2020-03-17
Feature registry

clelland:
This makes features.md the official registry for features, in accordance with W3C procedures for registries.

Closes: #244 , #75

### DIFF
--- a/features.md
+++ b/features.md
@@ -1,13 +1,24 @@
-# Policy Controlled Features
+# Feature Policy Feature Registry
 
-This document lists policy-controlled features being implemented in browsers. It
-is broken into sections based on the standardization state of each feature. The
-names used here should be consistent between browsers implementing a particular
-feature, but there is no requirement that all browsers implement any single
-feature.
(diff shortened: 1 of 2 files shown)


clelland:
@annevk , @domenic -- does this seem like a reasonable start to an actual feature registry?

clelland:
(The rest of the updates to the actual values in features.md are still coming, and I intend to have a very similar registry for document policy, but this is the starting point for sanity-checking the idea)

domenic:
I have no experience with registries, and am not a huge fan of them, so I'll leave this to others.

annevk:
There was an issue about potentially merging Feature Policy and Permissions that I don't think concluded yet. Should we first try to conclude that?

Malvoz:
@clelland As a reminder, [integration.md](https://github.com/w3c/webappsec-feature-policy/blob/master/integration.md#referencing-the-feature-policy-spec) currently states "There is no central registry for policy-controlled features".

Issue #376, opened 2020-05-11
Feature Policy syntax: Structured Header?


clelland:
If we [rename the header](https://github.com/w3c/webappsec-feature-policy/issues/359), then we have a chance to change the syntax, as existing sites would need to make changes anyway.

A concrete proposal here would be a dictionary, mapping features to lists of origins / keywords:

```http
Something-Policy: feature=(self "https://example.com"),
                  another-feature=(),
                  grant-it-to-everyone=(*)
```

Similar to ES6 arrow function syntax, we could choose to drop the parentheses if there is exactly one item in the list, for developer convenience. (Essentially allowing a single Item to take the place of an Inner List, in SH terms)

If we do this, what should we do with the existing `allow` attribute? I'd prefer to keep the existing name, and it's not clear that SH-syntax works well in this case -- the double-quotes around origins could be problematic, and there's no clear backwards-compatibility story. The ability to just use `<iframe allow="feature; another-feature; a-third-feature">` is a very useful (and widespread) shorthand, but there's no way to make that SH-compatible.

One option is to keep the existing syntax for the attribute, and define it as an alternate serialization of a policy declaration for use in HTML attributes, while the SH serialization is used for headers. I'm not sure if there's precedent for that.

It could also be that the benefits of having exactly one serialization outweigh all of my concerns, and we should just change everything all at once. I'm open to any and all suggestions here.

annevk:
I was thinking that we keep the `allow` attribute as-is as it has shipped in at least two implementations (not sure about Safari). It's unfortunate that it has a unique parser, but at least that parser would be bound to the content process.

clelland:
I think I'm happiest with that outcome, too. It's the least disruptive to developers.

Safari (WebKit at least, not sure about the shipping status) also supports the current allow syntax, so it's been implemented in three engines.

arturjanc:
My opinion carries little weight here, but I'd be happy if the new header followed the Structured Header format.

No opinions re: the `allow` attribute, though it does seem like having a different parser there could be a bit counterintuitive. Would it make sense to consider a different attribute (`permissions`?) with a syntax that matches SH, or is that even less understandable?

annevk:
It might be interesting to consider if there's a reasonable path to full migration, but as I understand it the `allow` attribute is already quite widely deployed. The next time we invent a complicated attribute for HTML we should strongly consider something like that though. I don't think it's acceptable to add another custom syntax at this point without a longer term plan.

[...]

PR #383, opened 2020-05-27
Define "Permissions-Policy" as a structured header

clelland:
Closes: #376

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -29,6 +29,9 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
   type: dfn
     text: report type
     text: visible to reportingobservers
+spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf-httpbis-header-structure.html#
+  type: dfn
+    text: sh-dictionary; url: dictionary
 </pre>
 <pre class="biblio">
 {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -80,10 +94,9 @@ spec:reporting; urlPrefix: https://w3c.github.io/reporting/
     within their application. It can do so by delivering the following HTTP
     response header to define a permissions policy:</p>
     <pre>
-      <a http-header>Permissions-Policy</a>: fullscreen 'none'; geolocation 'none'</pre>
-    <p>By specifying the "<code>'none'</code>"keyword for the origin list, the
-    specified features will be disabled for all browsing contexts, regardless of
-    their origin.</p>
+      <a http-header>Permissions-Policy</a>: fullscreen=(), geolocation=()</pre>
+    <p>By specifying an empty origin list, the specified features will be
+    disabled for all browsing contexts, regardless of their origin.</p>
```
annevk, line 99: documents (browsing contexts don't have an origin)

index.bs
```diff
@@ -743,14 +782,31 @@ partial interface HTMLIFrameElement {
     <div class="algorithm" data-algorithm="parse-header">
     Given a string (|value|) and an [=origin=] (|origin|), this algorithm will
     return a <a>declared permissions policy</a>.
+    1. Let |parsed header| be the result of parsing |policy| as a dictionary.
```
annevk, line 785: Where does policy come from? Also, I think this algorithm should take a SH dictionary. And elsewhere we'd use https://fetch.spec.whatwg.org/#concept-header-list-get-structured-header to get it from a header list.

clelland: Oops -- |policy| should have been |value| here.

And we can take a dictionary and defer to fetch for the SH parsing, thanks.

index.bs
clelland: Fixed, thanks.

PR #385, opened 2020-06-16
Add clipboard-read and clipboard-write Feature Policies

dway123:
The clipboard specification specifies permissions "clipboard-read" and "clipboard-write", and feature policy is used to control use of this API in iframes. As the clipboard specification gains feature policy [integration](https://github.com/w3c/clipboard-apis/pull/120), feature policy documentation should be updated to reflect this.

### DIFF
--- a/features.md
+++ b/features.md
@@ -21,6 +21,8 @@ specification.
 | `autoplay` | [HTML][html] | [Chrome 64](https://www.chromestatus.com/feature/5100524789563392) |
 | `battery` | [Battery Status API][battery-status] | Status "[Open](https://bugs.chromium.org/p/chromium/issues/detail?id=1007264)" in Chrome |
 | `camera` | [Media Capture][media-capture] | [Chrome 64](https://www.chromestatus.com/feature/5023919287304192) |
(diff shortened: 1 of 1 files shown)


annevk:
Are multiple implementers interested in this?

clelland:
There is ongoing discussion at https://github.com/w3c/permissions/issues/175 about this as well.

My understanding is that the Permissions API was changed from "clipboard" to "clipboard-read" and "clipboard-write", which Chrome implements. "clipboard-read" may not have interest from other implementers, and could be removed from the Permissions spec if so, but it looks like "clipboard-write" does.

If the goal of features.md is to document what is in specs and what browsers support, then we should use it for that (even if there is only a single-implementation, we can at least document that). I'd suggest waiting a few days to see how that issue is resolved to know where the two feature names should go.

clelland:
This is still unresolved in the spec, but is implemented in Chromium; I'll add it to the 'proposed' section for now, rather than merging this in. Please feel free to reopen or just ping when if and when the integration is merged into Clipboard

PR #394, opened 2020-07-16
web-share permission policy

ewilligers:
"web-share" was recently added to the Web Share API spec.
w3c/web-share#151

Blink support is being implemented in
https://chromium-review.googlesource.com/c/chromium/src/+/2299791


ewilligers:
@clelland  please review

clelland:
Looks good, thanks!

PR #400, opened 2020-08-17
Update all links to webappsec-permissions-policy

tidoust:
The specification's shortname was updated from `webappsec-feature-policy` to `webappsec-permissions-policy` but some links (including the "This version" link) still used the previous shortname.

This update replaces all occurrences of `webappsec-feature-policy` with `webappsec-permissions-policy` to avoid redirects.


clelland (approved):
This is all good, thanks!

PR #405, opened 2020-09-24
Alphabetize list of features

dbaron:
This was mostly alphabetical, but a few were not quite in order.


clelland:

clelland:
Merged, thanks!

PR #407, opened 2020-10-05
s/feature-policy/permissions-policy in urls

tantek:


clelland:
Thanks for fixing those!

Issue #408, opened 2020-10-09
Header should be sufficient in some cases to delegate features (esp. Client Hints)


clelland:
I think that the changes made in #378 may have gone a bit further than necessary. #357, which was the motivation for the change, asked that blocking policies set in the header not be overridable with the allow attribute.

#378 does this, by making policy inheritance essentially the logical AND of parent-header and iframe attribute policy, and necessarily switches the header defaults for every feature to "*". (That was necessary so that developers would not be *required* to set the header in order to enable delegation at all, but could use just the attribute in the common case.) This also ends up making the allow attribute required in every case to delegate any permission.

Unfortunately, this model breaks some deployed use cases, where the header could previously have been used to explicitly grant permission to specific origins, and isn't very ergonomic for the Client-hint delegation case, where not only would every hint need to be specified on every frame, but we lose the ability to tell from the header which hints should be sent with subresource requests.

@yoavweiss suggested in email a small tweak which would reenable the use case of "enable delegation for specific sites with the header" and would allow Client Hints to be delegated with the header only. It involves treating an explicit header declaration, where a feature is mentioned in the header, with an associated allowlist, differently from the 'default' behaviour when the feature is not mentioned at all in a header.

> Thinking about this a bit, it seems to me that the current state of the `allow` attribute when it comes to Client Hints and iframe delegation is not something that most developers would be able to use.
>
> Currently developers are required to add an `Accept-CH` header in order to be able to delegate Client Hints (which already multiple folks found confusing). If we add a Permissions Policy requirement on top of that, that is not likely to make things better.
>
> The way most people delegate permissions is by providing developers with a snippet of HTML they can integrate into their site, including a list of `allow` values. Asking them to muck around with headers is completely orthogonal to this flow, and doesn't serve much when it comes to the goal of CH being "accountable". The browser can keep track of an allow attribute in exactly the same way it can keep track of a header-based opt-in.
>
> That that end, I propose:
> * To change Client Hints' delegation model so that an Accept-CH opt-in will not be required in order to be able to delegate hints to a header.
> * To change Permission Policy so that explicit restrictions are different than default restrictions
>     * We want explicit restrictions to not be overridable, but implicit (by default) restrictions should be overridable.
>     * That way, CH are handled correctly, without special-casing them.

I want to propose a change which would allow this, as well as the existing use cases, while still being web-compatible:

* Header allowlist is 'default' unless otherwise specified. Iframe allowlist is 'default' unless otherwise specified.
[...]

PR #409, opened 2020-10-09
Allow header to enable features.

clelland:
This allows the header alone to enable delegation of a feature to specific
origins, if not otherwise blocked or affected by the container policy. The
order of precedence becomes:

1. Explicitly blocked by header: Disabled
2. Explicitly blocked by allow attribute: Disabled
3. Explicitly allowed by allow attribute: Enabled
4. Explicitly allowed by header: Enabled
5a. (Default behaviour when default allowlist is '*'): Enabled
5b. (Default behaviour when default allowlist is 'self'): Enabled if same-
    origin; Disabled if cross-origin.

Fixes: #408

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -932,6 +932,9 @@ partial interface HTMLIFrameElement {
         1. If the <a>allowlist</a> for |feature| in |container policy|
           <a>matches</a> |origin|, return "<code>Enabled</code>".
         1. Otherwise return "<code>Disabled</code>".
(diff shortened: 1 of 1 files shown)


yoavweiss (commented):
non-owner LGTM

clelland:
Closing this; the resolution of #408 is that we're not going to make this change to the algorithms right now.

PR #416, opened 2020-12-10
Remove document policy from this repo.

### LINKED ISSUE #411: Move Document Policy
As mentioned in passing in https://github.com/whatwg/xhr/pull/295 I think it would be better for it to be in its own repository. Also, as it's a new thing it should probably be incubated first?

clelland:
This removes the document policy spec and associated machinery from this repository,
and updates references to point to the new location at WICG. A redirect is put in
place from the old spec location to the new one.

Closes: #411


annevk (approved):

annevk:
(This looked okay, but Travis seems to disagree I see now...)

clelland:
I put in a static file to redirect from <here>/document-policy.html to the new repo location, since that's something that GitHub won't automatically do, and there may be existing references to that location.

Or at least I attempted to do that. I didn't include the file with the commit, so Travis was understandably confused.

matrawybarte-lgtm (commented):
Kita

PR #417, opened 2021-02-20
Add .circleci/config.yml

Brenoamandobg:


clelland:
I'm not sure what this is for; this appears to be a continuous integration product that we don't use?
I'm closing this, but if you can give more context, or a bug justifying the change, then that may be something to reconsider.

PR #425, opened 2021-05-12
Updated spec links to updated hash values

yoshifp:


clelland:
Thanks for updating these -- I've validated them and am merging this now.

PR #431, opened 2021-10-20
Default allowlists only effect child browsing contexts

reillyeon:
This change updates the definition of a "default allowlist" to clarify
that it only effects feature availability in child browsing contexts.
That is, a feature is always available by default in a top-level
browsing context but may be specified with a default allowlist of '*' or
'self' in order to control whether the feature is available by default
in cross-origin documents in child browsing contexts.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -308,22 +308,19 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
     <h3 id="default-allowlists">Default Allowlists</h3>
     <p>Every <a>policy-controlled feature</a> has a <dfn export lt=
     "default allowlist|default allowlists">default allowlist</dfn>. The
-    <a>default allowlist</a> determines whether the feature is allowed in a
-    document with no declared policy in a top-level browsing context, and also
-    whether access to the feature is automatically delegated to documents in
-    child browsing contexts.</p>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -308,22 +308,19 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      default, and when allowed, is allowed by default to documents in child
-      browsing contexts.</dd>
+      <dd>When allowed in the top-level document, the feature is delegated by
+      default to documents in child browsing contexts.</dd>
       <dt><code>'self'</code></dt>
-      <dd>The feature is allowed in documents in top-level browsing contexts by
-      default, and when allowed, is allowed by default to same-origin domain
-      documents in child browsing contexts, but is disallowed by default in
-      cross-origin documents in child browsing contexts.</dd>
+      <dd>When allowed in the top-level document, the feature is delegated by
```
clelland, line 320: This process applies recursively down the frame tree as well -- if the feature is allowed in some embedded frame, then it will be delegated by default to *its* same-origin children. I'd like to find a way to express that here, since it's not just about the top-level.

clelland:
Closing this; the text around here has changed significantly. Some of this might still be useful, but would need to be rewritten before I could merge it.

PR #438, opened 2021-11-29
add experimental features join-ad-interest-group and run-ad-auction

qingxinwu:
Features join-ad-interest-group and run-ad-auction are enabled by default for all frames. The default behaviour will be changed to same-origin iframes only in near future, before third party cookie is deprecated.
There is a flag AdInterestGroupAPIRestrictedPolicyByDefault (--enable-features=AdInterestGroupAPIRestrictedPolicyByDefault) which will disable the two features for cross-origin iframes by default. Then cross-origin iframes need to specify an allow attribute on the iframe, e.g. <iframe src="some_page.html" allow="join-ad-interest-group"></iframe>, in order to access the features.

### DIFF
--- a/features.md
+++ b/features.md
@@ -67,6 +67,8 @@ experimentation by web developers.
 | `focus-without-user-activation` | [focus-without-user-activation.md](policies/focus-without-user-activation.md) | Status "[Open](https://bugs.chromium.org/p/chromium/issues/detail?id=965495)" in Chrome |
 | `hid` | [Explainer](https://github.com/WICG/webhid/blob/master/EXPLAINER.md) | In [Origin Trial](https://developers.chrome.com/origintrials/#/view_trial/1074108511127863297) in Chrome 86-88 |
 | `idle-detection` | [Draft spec](https://github.com/WICG/idle-detection/pull/29) | In [Origin Trial](https://developers.chrome.com/origintrials/#/view_trial/551690954352885761) in Chrome 84-86 |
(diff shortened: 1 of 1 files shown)


mikewest:
Would you be willing to put up a PR against https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md listing these features and linking to documentation explaining their function?

mikewest:
(Also, for IPR reasons, you'll need to link your GitHub account with a W3C account, and follow the (Google internal) instructions at go/github to get into the right GitHub group.)

qingxinwu:
The explainer is updated to mention the feature names. It also talks about the ideal default allowlist, but we think the  allowlist details fit more in a spec (which is still in progress) than in the explainer. Please let me know if there's any question. Thanks!

qingxinwu:
Hi @mikewest, we're writing the [spec](https://github.com/WICG/turtledove/pull/366) for FLEDGE, which needs to document the Permissions policy requested in this PR.
Would you mind reviewing this PR again? Thanks!

clelland:
Hey @qingxinwu - would it makes sense at this point to link these features to the spec instead, now that the feature definition has been integrated there?

qingxinwu:
> Hey @qingxinwu - would it makes sense at this point to link these features to the spec instead, now that the feature definition has been integrated there?

Updated to link to the spec.
Not sure why I missed this comment, sorry about that.

clelland:
Thanks!

PR #447, opened 2022-03-24
Replace usage of "responsible document"

johannhof:
See https://github.com/whatwg/html/pull/7694

### DIFF
--- a/integration.md
+++ b/integration.md
@@ -27,7 +27,8 @@ or not in a specific document, with text similar to this:
 
 > Example:
 >
-> If the [responsible document](https://html.spec.whatwg.org/multipage/webappapis.html#responsible-document)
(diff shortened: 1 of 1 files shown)


johannhof:
@domenic can you check if I'm using the current global object correctly in this example? Thanks :)

domenic (commented):
Usually you want the relevant global object of some object (such as [this](https://webidl.spec.whatwg.org/#this)). See https://html.spec.whatwg.org/multipage/webappapis.html#:~:text=In%20general%2C%20web%20platform%20specifications%20should%20use%20the%20relevant%20concept

johannhof:
Makes sense, thanks! Updated!

domenic (approved):

PR #451, opened 2022-04-13
Export all algorithms (as algorithms)

eeeps:
As discussed in https://github.com/WICG/client-hints-infrastructure/pull/107#issuecomment-1098002600

A couple of these were already exported, but not as algorithms. I assume this change will break incoming links; is there an easy way to see existing incoming links?

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -741,7 +741,7 @@ partial interface HTMLIFrameElement {
 <section>
   <h2 id="algorithms">Algorithms</h2>
   <section>
-    ## <dfn export>Process response policy</dfn> ## {#algo-process-response-policy}
+    ## <dfn export abstract-op id="process-response-policy">Process response policy</dfn> ## {#algo-process-response-policy}
 
     <div class="algorithm" data-algorithm="process-response-policy">
     Given a [=response=] (|response|) and an [=origin=] (|origin|), this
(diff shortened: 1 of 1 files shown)


clelland:
Unfortunately not; I think there is work being done on a spec cross-references database, but I don't think anything is available yet.

Thanks for doing this! I'll check it out and make sure the rendered HTML is okay, and merge it.

johannhof:
Maybe I misunderstand what you're trying to do, but does https://dontcallmedom.github.io/webdex/ help? :)

clelland:
That's exactly what I was thinking of, thank you! Now to see if it can find this, or just already-properly-exported definitions

clelland:
It doesn't look like it breaks links to actual fragments -- anything which was linking to `#algo-foo` will still work; the ids haven't changed.

Internal links that just used `<a>` seem to be broken by this, though. (See, for example, step 3 in algorithm 9.1: It reads "Let policy be the result of executing Construct policy from dictionary and origin on parsed header and origin.", and "Construct policy from dictionary and origin" should be linked to algorithm 9.2)

I think the solution here is to change the `<a>` in that case to `<a abstract-op>`; that seems to restore the link in the rendered output.

eeeps:
Ok. I think I got them all @clelland, but please review!

eeeps:
Note that the original reason for this, https://github.com/WICG/client-hints-infrastructure/pull/107, no longer, well, needs this; sounds like `<meta name=accept-ch>` will use its own new syntax/parsing algorithm.

It still might be a good idea to export all of the Permissions Policy algorithms, and it definitely seems like a good idea to classify them as abstract-ops rather than generic dfns.

clelland:
Agreed -- and I think you got all of the references; I didn't see any go missing when I re-rendered the output.
Merging this now, thanks!

PR #458, opened 2022-04-26
Add local-fonts permission policy

inexorabletash:


freddieleeman:
This should not be placed in the "Standardized Features" table but in the "Experimental Features" table.

inexorabletash:
Happy to move it, but the tables are currently a bit confusing.

* **Standardized Features** contains several items that are marked "Experimental in Chrome xx" or "Behind a flag in Chrome", and the table header reads _"These features have been declared in a published version of the respective specification."_ 
* **Proposed Features** is described as _"These features have been proposed, but the definitions have not yet been integrated into their respective specs"_ - which doesn't obviously apply, as the spec does have a definition integrated.
* **Experimental Features** is described as _"These features generally have an explainer only..."_ which allows anything, but all the examples given  link to an explainer not a spec, even though some of them have specs (and need updating, e.g. "window-placement" which shipped in Chrome 100) 

I think the first table, "Standardized Features", is conflating _specified_ (i.e. there is a formal spec) with _standardized_ (i.e. implemented by multiple browsers and/or adopted by a working group of some sort). 

Aaaaaaanyway, happy to update if Experimental Features is indeed where it should land. Would you want the link to be to the spec and explainer?

clelland:
Resolving merge conflicts and committing here.

Issue #471, opened 2022-05-23
New permission policy features that should get a mention in this repo


summercms:
The following page: https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md

Should get updated to mention the following:

### Standardized Features

| Feature name | Spec link(s) | Browser Support |
| ------------ | ------------ | --------------- |
| `bluetooth` | https://webbluetoothcg.github.io/web-bluetooth/#permissions-policy | [Chrome 104](https://chromestatus.com/feature/6439287120723968) |
| `ch-prefers-color-scheme` | https://wicg.github.io/user-preference-media-features-headers/#policy-controlled-features | [Chrome 93](https://chromestatus.com/feature/5642300464037888) |

### Proposed Features

| Feature name | Spec link(s) | Browser Support |
| ------------ | ------------ | --------------- |
| `ch-prefers-reduced-motion` | https://wicg.github.io/user-preference-media-features-headers/#policy-controlled-features | n/a |
| `ch-prefers-reduced-transparency` | https://wicg.github.io/user-preference-media-features-headers/#policy-controlled-features | n/a |
| `ch-prefers-contrast` | https://wicg.github.io/user-preference-media-features-headers/#policy-controlled-features | n/a |
| `ch-prefers-forced-colors` | https://wicg.github.io/user-preference-media-features-headers/#policy-controlled-features | n/a |
| `direct-sockets` | https://wicg.github.io/direct-sockets/#permissions-policy | n/a |
| `request-storage-access` | https://privacycg.github.io/storage-access/#permissions-policy-integration | Shipped: Firefox and Safari but not [Chrome](https://chromestatus.com/feature/5612590694662144) |

Note: `direct-sockets` is in Chromium source code found here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/permissions_policy/permissions_policy_features.json5;l=212-215

Note: `request-storage-access` is under the name `storage-access-api` in Chromium source code found here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/permissions_policy/permissions_policy_features.json5;l=349-352 opened an issue to find out: https://github.com/privacycg/storage-access/issues/104

### Experimental Features

| Feature name | Link(s) | Browser Support |
| ------------ | ------- | --------------- |
| `sync-script` | | Behind a flag in Chrome<sup>[1](#fn1)</sup> |

[...]

PR #473, opened 2022-05-24
Add new permission policy features

summercms:
Github issue: https://github.com/w3c/webappsec-permissions-policy/issues/471

Have updated the `features.md` file as it was very out of date!

### Deprecated Features

Added a `Deprecated Features` section, features that were added to a browser and are still listed in the browsers source code, but have since been deprecated or replaced with a different feature name.

### Behind a flag (Chromium)

As of May 2002, see here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/permissions_policy/permissions_policy_features.json5

- `ch-ua-platform`
- `join-ad-interest-group`
- `run-ad-auction`

Are behind flags according to the Chromium source code and so updating the doc's accordingly.

### DIFF
--- a/features.md
+++ b/features.md
@@ -73,6 +73,7 @@ experimentation by web developers.
 
 | Feature name | Link(s) | Browser Support |
 | ------------ | ------- | --------------- |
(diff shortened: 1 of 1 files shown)


summercms:
IPR:

> [summercms](https://github.com/summercms/) did not make IPR commitments for this group. To make the IPR commitments, summercms should [join the Web Application Security Working Group](https://www.w3.org/groups/wg/webappsec/join) .

> You cannot join the group. 

How to sort out the IPR ?

PR #476, opened 2022-06-22
Add Bluetooth to the list of policy-controlled features

gabrielsanbrito:
Hi!
We have integrated recently Web Bluetooth with Permissions Policy. We would like to update this table with the related info.


clelland:
That's great! Merged.

PR #482, opened 2022-08-17
Wildcards in Permissions Policy Origins

### LINKED ISSUE #479: Client Hint delegation to multiple subdomains
Hi everyone!

As part of the User Agent Reduction initiative (in Chrome), we (Akamai mPulse RUM) are implementing some changes for our product to request the information that will soon be frozen/removed from the UA string via Client Hints instead.  For example, we would like to continue being able to classify RUM visitors by their Platform Version (and not just Platform, which is included as a default CH header).

For our customers, the Akamai CDN (or their origins) will be injecting the necessary `Accept-CH` headers, e.g:

```
accept-ch: Sec-CH-UA-Platform-Version
```

Our product interacts with a few domains that we would like the Platform Version sent to, such as:

* `c.go-mpulse.net` to fetch a configuration file
* `*.akstat.io` to send the RUM beacon

So we were planning on including a `Permissions-Policy` header to delegate the requested Client Hints to those domains.  For example, we were expecting to be able to send something like:

```
permissions-policy: ch-ua-platform-version=(self "https://*.akstat.io" "https://c.go-mpulse.net") 
```

This works fine for `c.go-mpulse.net` but our request of a wildcard in `*.akstat.io` does not work.

Our `*.akstat.io` domains receiving beacons have 100s+ subdomains mapping to specific paths we want the beacon to take, i.e. it may be `abcd1234.akstat.io` for one visitor and `efgh5678.akstat.io` for another.  We don't know the domain at the time the HTML page is sent to the browser.

We were hoping that a wildcard (subdomain) match for `*.akstat.io` in the `Permissions-Policy` would work, but that does not seem to be the case (and isn't spec'd as such).

Without making any changes to the spec, we brainstormed a few options to resolve this:

1. Delegate to all subdomains: `ch-ua-platform-version=(self "*")`
    * This is not ideal, as it sends whatever Client Hints we're requesting to other 3P domains that aren't requesting it
    * This will take additional bandwidth on all other 3P requests
    * For the sake of privacy, we would like to avoid doing this.
2. Explicitly list all of the potential subdomains: `ch-ua-platform-version=(self "https://abcd1234.akastat.io" "https://efgh5678.akastat.io" ...)`
    * We have 100s of potential subdomains, so maintaining this list would be a challenge
    * It would waste a lot of bandwidth
3. Use the JavaScript high-entropy API to get this data instead (in boomerang.js, the JS library that our customers use)
    * We are planning on doing this, but our customers are slow/hesitant to upgrade their boomerang.js versions so it may be years before they do so

Instead, we are hoping a proposal to allow for subdomains to match PP would be feasible and secure.

There was a bit of previous discussion on wildcards and subdomains here:

https://github.com/w3c/webappsec-permissions-policy/issues/129

And even a note from @clelland:

> CSP allows much richer wildcards than feature policy currently does -- If FP adopted some of the same parsing logic, then you could cut it down to

But as far as I can tell, wildcards/subdomains didn't make it into the final Client Hints spec.

I understand that "star" wildcards (e.g. `*.akstat.io`) may have security concerns, but we were hoping some sort of syntax that might allow for any subdomain of an origin to get data, maybe something like this amateur attempt at a syntax:

```
permissions-policy: ch-ua-platform-version=(self "subdomains:https://akstat.io" "https://c.go-mpulse.net") 
```

Anyways, we're just trying to do the right thing and not request these hints and blast them all over!

Would love your thoughts,

arichiv:
This change will add support for wildcard in permissions policy structured like SCHEME://*.HOST:PORT (e.g., https://*.foo.com/) where a valid Origin could be constructed from SCHEME://HOST:PORT (e.g., https://foo.com/). This requires that HOST is at least eTLD+1 (a registrable domain). This means that https://*.bar.foo.com/ works but https://*.com/ won’t (if you want to allow all domains to use the feature, you should just delegate to *). Wildcards in the scheme and port section will be unsupported and https://*.foo.com/ does not delegate to https://foo.com/.

Before, a permissions policy might need to look like:
permissions-policy: ch-ua-platform-version=(self "https://foo.com" "https://cdn1.foo.com" "https://cdn2.foo.com") 

With this feature, it could look like:
permissions-policy: ch-ua-platform-version=(self "https://foo.com" "https://*.foo.com") 

This flexibility assists developers that need to delegate permissions to multiple subdomains (e.g., CDNs).

This is an in-progress change, the following components are needed:
- [x] Update ABNF
- [x] Define matching
- [x] Add example

Design Doc: https://docs.google.com/document/d/1_Y5qRs0CO2_R7o8M_YaVAWQvyjXniG70RG2z1a2uwN8/edit

closes #479

***
<a href="https://pr-preview.s3.amazonaws.com/arichiv/webappsec-permissions-policy/pull/482.html" title="Last updated on Oct 5, 2022, 2:57 PM UTC (7f4572c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/482/432a153...arichiv:7f4572c.html" title="Last updated on Oct 5, 2022, 2:57 PM UTC (7f4572c)">Diff</a>


arichiv:
@yoavweiss have a sec for review? I want to get the ABNF correct before moving on to deeper integration

The other option is to do something more explicit like https://www.w3.org/TR/CSP2/#host_part, but it seemed desirable to not reinvent the wheel since we aren't supporting wildcards everywhere.

arichiv:
Forgot @yoavweiss is out, @clelland have time to take a look?

index.bs
```diff
@@ -338,13 +338,16 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / "*" / "'self'" / "'src'" / "'none'"
+      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / <a>serialized-origin-with-wildcard-subdomain</a> / "*" / "'self'" / "'src'" / "'none'"
     </pre>
     <p><dfn><code>serialized-origin</code></dfn> is the
     <a>serialization of an origin</a>. However, the code points U+0027 ('),
     U+0021 (*), U+002C (,) and U+003B (;) MUST NOT appear in the serialization.
     If they are required, they must be percent-encoded as "`%27`", "`%2A`",
     "`%2C`" or "`%3B`", respectively.</p>
+    <p><dfn><code>serialized-origin-with-wildcard-subdomai</code></dfn> is a
+    <a>serialized-origin</a> with (1) a <a>registrable domain</a> for the <a>host</a>
```
clelland, line 349: I think it's problematic to say that a `serialized-origin-with-wildcard` *is a* `serialized-origin`, when it explicitly breaks some of the conditions that a `serialized-origin` has (namely, that `*` cannot appear in it.)

Maybe we can say that it is the serialization of an origin whose host is a domain which is the concatenation of the string "*." and a domain which has a non-null registrable domain.

That's a little bit odd, since a domain isn't really supposed to have a '*' in it, but it's probably the best we can do without going the CSP route and defining a full syntax for it.

I don't think you want to require the remainder after the "*." to be a registrable domain on its own, since that would preclude things like `*.staging.mywebsite.example` -- in that case, `staging.mywebsite.example` isn't a registrable domain (`mywebsite.example` is).

arichiv:
Updated with the matching logic.

index.bs
```diff
@@ -297,8 +297,28 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
       <li>Otherwise, for each <var>item</var> in the <a>allowlist</a>:
         <ol>
-          <li>If <var>item</var> is [=same origin-domain=] with
-          <var>origin</var>, then return true.</li>
+          <li>If <var>item</var> is a <a>serialized-origin</a> that's [=same origin-domain=] 
+          with <var>origin</var>, then return true.</li>
+          <li>Else if <var>item</var> is a <a>serialized-origin-with-wildcard-subdomain</a>
+          then:</li>
+          <ol>
+            <li>If <var>origin</var> does not have a host with a <a>registrable domain</a>
```
yoavweiss, line 305: Can you link "host" as well? "does not have" is also a bit ambigious, and it might be better to link origin to https://html.spec.whatwg.org/multipage/origin.html#concept-origin and indicate that the host is not part of the tuple.

yoavweiss: Also, the registrable domain algorithm can return null, but there's no concept of a host without a registrable domain..

index.bs
```diff
@@ -338,13 +358,16 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / "*" / "'self'" / "'src'" / "'none'"
+      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / <a>serialized-origin-with-wildcard-subdomain</a> / "*" / "'self'" / "'src'" / "'none'"
     </pre>
     <p><dfn><code>serialized-origin</code></dfn> is the
     <a>serialization of an origin</a>. However, the code points U+0027 ('),
     U+0021 (*), U+002C (,) and U+003B (;) MUST NOT appear in the serialization.
     If they are required, they must be percent-encoded as "`%27`", "`%2A`",
     "`%2C`" or "`%3B`", respectively.</p>
+    <p><dfn><code>serialized-origin-with-wildcard-subdomain</code></dfn> is a serialization of an
+    <a>origin</a> whose host is a domain which is the concatenation of the string "*." and a
```
yoavweiss, line 369: Maybe 'is a serialization of a |tuple origin| whose |tuple|'s |host| is a |domain|, has a non-null |registrable domain|, and is |prefixed| by the string "*."'?

Also: "spec:html; type:dfn; for:/; text:origin" at the top of the spec should be removed, as origin is linking to the wrong concept.

index.bs
```diff
@@ -297,8 +297,28 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-          <li>If <var>item</var> is [=same origin-domain=] with
-          <var>origin</var>, then return true.</li>
+          <li>If <var>item</var> is a <a>serialized-origin</a> that's [=same origin-domain=] 
+          with <var>origin</var>, then return true.</li>
+          <li>Else if <var>item</var> is a <a>serialized-origin-with-wildcard-subdomain</a>
+          then:</li>
+          <ol>
+            <li>If <var>origin</var> does not have a host with a <a>registrable domain</a>
+            return false.<li>
+            <li>Set <var>originWithoutWildcard</var> to be <var>item</var> without the "*."
```
yoavweiss, line 307: Maybe use code point substring here? https://infra.spec.whatwg.org/#code-point-substring

arichiv: I dunno, that seems overly complicated, is the updated language okay? I wanted to mirror what https://infra.spec.whatwg.org/#strip-and-collapse-ascii-whitespace did

index.bs
```diff
@@ -297,8 +317,28 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+          then:</li>
+          <ol>
+            <li>If the <var>tuple origin</var> lacks a <a>host</a> with a <a>registrable domain</a>
+            return false.<li>
+            <li>Set <var>originWithoutWildcard</var> to be equal to <var>item</var> with the sequence
+            of <a>code points</a> "*." removed at the start of the <a>host</a>.</li>
+            <li>If <var>originWithoutWildcard</var> is [=same origin-domain=] </li> with
+            <var>origin</var>, then return false.</li>
+            <li>Set <var>originCandidate</var> to be equal to <var>origin</var> with the
+            left most domain label (and the dot after it) removed.</li>
```
yoavweiss, line 332: I think you'd need stricter definition of what"left most domain label removed" is. (that is, loop over the code points until you find a dot, and substring from that point, or something similar)

index.bs
```diff
@@ -338,13 +378,16 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
       <dfn>allow-list</dfn> = <a>allow-list-value</a> *(RWS <a>allow-list-value</a>)
-      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / "*" / "'self'" / "'src'" / "'none'"
+      <dfn>allow-list-value</dfn> = <a>serialized-origin</a> / <a>serialized-origin-with-wildcard-subdomain</a> / "*" / "'self'" / "'src'" / "'none'"
     </pre>
     <p><dfn><code>serialized-origin</code></dfn> is the
     <a>serialization of an origin</a>. However, the code points U+0027 ('),
     U+0021 (*), U+002C (,) and U+003B (;) MUST NOT appear in the serialization.
     If they are required, they must be percent-encoded as "`%27`", "`%2A`",
     "`%2C`" or "`%3B`", respectively.</p>
+    <p><dfn><code>serialized-origin-with-wildcard-subdomain</code></dfn> is a serialization of an
```
yoavweiss, line 388: s/an/a/

index.bs
```diff
@@ -338,13 +378,16 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+    <a>tuple origin</a> whose <a>tuple</a>'s <a>host</a> is a <a>domain</a> has a non-null
```
yoavweiss, line 389: ", " before "has"?

index.bs
yoavweiss: OK, current language seems fine

yoavweiss (commented):
Non-authoritative LGTM

clelland:
I think there's still a piece missing here -- we're going to need X parts to incorporate this:
 - The syntax of an origin with a wildcard (which I think is good)
 - An internal representation that can be part of an allowlist
 - The parsing logic to get from a header string to that internal representation
 - The matching logic to compare that internal representation to an origin in algorithms

So far, this has the syntax, and correctly defines serialized-origin-with-wildcard-subdomain as a string, but those strings aren't allowed in allowlists (per https://w3c.github.io/webappsec-permissions-policy/#allowlists), and I think that the matching logic is trying to operate directly on those strings, which is awkward -- and to do it right, we'd need to convert back and forth between strings and origins.

A way to resolve this would be to change the definition of what an allowlist can contain -- define an "origin with possible wildcard" to be an origin, and a boolean that says whether it also matches subdomains. Then the matching logic can check that boolean to know whether to do the deeper match.

Regardless of how we handle it, we'll need to change [9.2. Construct policy from dictionary and origin](https://w3c.github.io/webappsec-permissions-policy/#algo-construct-policy) and [9.3. Parse policy directive](https://w3c.github.io/webappsec-permissions-policy/#algo-parse-policy-directive) to parse the serialized-origin-with-wildcard-subdomain strings explicitly. Currently the call to the URL parser will fail because of the wildcards.

arichiv:
@clelland good point! I think I integrated at the necessary spots now.

index.bs
```diff
@@ -297,8 +318,25 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
       then return true.</li>
       <li>Otherwise, for each <var>item</var> in the <a>allowlist</a>:
         <ol>
-          <li>If <var>item</var> is [=same origin-domain=] with
-          <var>origin</var>, then return true.</li>
+          <li>If <var>item</var> is a tuple of a <a>serialized-origin</a> that's [=same origin-domain=]
```
clelland, line 321: You probably just want "an [=origin=]" rather than a serialized-origin, since allowlists are now defined to contain (origin,bool) tuples (same on line 323)

index.bs
```diff
@@ -775,16 +816,17 @@ partial interface HTMLIFrameElement {
           <code>*</code></a>.
         1. Otherwise:
             1. Set |allowlist| to an new <a>ordered set</a>.
-            1. If |value| is the token `self`, append |origin| to |allowlist|.
+            1. If |value| is the token `self`, append the tuple |origin|, <code>False</code> to |allowlist|.
```
clelland, line 819: You can use [tuple literal syntax](https://infra.spec.whatwg.org/#tuples) here -- just surround `origin, false` with parentheses

index.bs
```diff
@@ -828,12 +870,14 @@ partial interface HTMLIFrameElement {
...
                     1. Let |target| be the origin of |result|.
-                    1. If |target| is not an opaque origin, append |target| to
+                    1. If |target| is not an opaque origin, append the tuple |target|, <code>False</code> to
                       |allowlist|.
         1. Set |directive|[|feature|] to |allowlist|.
     1. Return |directive|
 
     </div>
+    
+    Note: This algorithm does not support wildcard subdomains in allowlists.
```
clelland, line 880: Is it deliberate that wildcards are not supported in the attribute syntax? (Just that if not -- if it's something to fix -- it should be an Issue rather than a note)

clelland (changes requested):

index.bs
arichiv: It is deliberate, I added another sentence to that effect

index.bs
```diff
@@ -297,8 +318,25 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
       then return true.</li>
       <li>Otherwise, for each <var>item</var> in the <a>allowlist</a>:
         <ol>
-          <li>If <var>item</var> is [=same origin-domain=] with
-          <var>origin</var>, then return true.</li>
+          <li>If <var>item</var> is a tuple of an [=origin=] that's [=same origin-domain=] 
+          with <var>origin</var> and the boolean <code>False</code>, then return true.</li>
+          <li>Else if <var>item</var> is a tuple of a <a>serialized-origin</a> that's not [=same origin-domain=] 
+          with <var>origin</var> and the boolean <code>True</code>, then:</li>
```
annevk, line 324: How can you compare a serialized-origin like that? Wouldn't it have to be parsed first?

index.bs
```diff
@@ -297,8 +318,25 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+          <ol>
+            <li>If <var>origin</var> is not a <a>tuple origin</a>, then return false.<li>
+            <li>If <var>origin</var>'s <a>host</a> has a null <a>registrable domain</a>, return false.<li>
```
annevk, line 327: This has to refer to the host of an origin using the `for` attribute.

index.bs
```diff
@@ -297,8 +318,25 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+            <li>Set <var>originWithoutWildcard</var> to be equal to the first value of the tuple <var>item</var>.</li>
+            <li>Set <var>originCandidate</var> to be equal to <var>origin</var> with the sequence
+            of <a>code points</a> in the <a>host</a> from the start up to and including the first "." removed.</li>
+            <li>While <var>originCandidate</var> has a <a>host</a> with a non-nill <a>registrable domain</a>:</li>
```
annevk, line 331: Same problem with the host field as above.

Also, it's non-null.

index.bs
```diff
@@ -297,8 +318,25 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+            <ol>
+              <li>If <var>originWithoutWildcard</var> is [=same origin-domain=] </li> with
```
annevk, line 333: That `</li>` seems misplaced.

annevk (commented):
I thought we reached agreement at TPAC that this should share logic with CSP?

arichiv:
> I thought we reached agreement at TPAC that this should share logic with CSP?

I thought the agreement was that any further expansion of the wildcard syntax would require sharing logic with CSP, but that for this restrictive wildcard support a local modification was acceptable.

index.bs
annevk: Could you explain how you resolved this, as far as I can tell you reordered the words, but that doesn't help the fact that host points to the wrong thing.

index.bs
annevk: Same here.

annevk:
I think that any kind of wildcard parsing needs to have shared logic. We don't want multiple independent code paths for this that need to be audited separately and have the opportunity for divergence if future maintainers are not careful.

index.bs
arichiv: I think I had misunderstood, is it correct now?

arichiv:
Restarting this since it's been a while. Will live on in:
https://github.com/w3c/webappsec-permissions-policy/pull/516

PR #486, opened 2022-09-15
Use spec-prod instead of Travis CI.

jyasskin:


jyasskin:
A repo admin might need to turn on GH Actions in order for this one to run.

marcoscaceres (approved):

marcoscaceres:
I imagine @sideshowbarker should have rights to enable it.

sideshowbarker:
https://w3c.github.io/webappsec-permissions-policy/ shows it was built from https://github.com/w3c/webappsec-permissions-policy/commit/432a1532c943a079351eecbc445ae4c6ed9b949c and https://github.com/w3c/webappsec-permissions-policy/actions/runs/3094677061/jobs/5008306993 is all green, so that’s all working.

But we didn’t yet also enable auto-publishing to https://www.w3.org/TR/permissions-policy/ — but there’s no reason we shouldn’t, so I’ll try to make time today or later this week to flip that on too.

clelland:
Thanks, @sideshowbarker! I'll set up an issue for that, just so that it doesn't get lost.

PR #488, opened 2022-09-23
Propose identity-credential-get permission policy

pkotwicz:


clelland:
@pkotwicz, do you mind updating this to point to the section in the spec that defines this (rather than the pull request)?

Thanks!

PR #496, opened 2022-12-08
Add smart-card

dandrader:


clelland:
Hey @dandrader, the linked explainer doesn't mention anything at all about permissions policy, (or policy-controlled features, or even 'policy',) as far as I can tell. Is there a better document to link to that would?

dandrader:
Hi Ian,
You're right. That will come with the spec as I thought it would be too much to have this in the explainer.

PS: don't know why I didn't get a notification email or something when you commented...

PR #499, opened 2022-12-15
Update "Is feature enabled in document for origin?" behavior

xyaoinum:
Rationale: if a page can create an <iframe src=… allow="featureX"> and can delegate permissions to the iframe, then the request like fetch(url, {allowFeatureX: true}) should be allowed for the permission as well, otherwise, they can just create an iframe similarly to gain the permission.

For a feature that does not have an equivalent opt-in flag, this rationale isn’t 100% applicable, so I’ll leave the spec unchanged for now. But for feature that does have it (e.g. fetch(url, {browsingTopics: true}) for the [Topics API](https://github.com/patcg-individual-drafts/topics)), we should be able to skip the last few steps that check the default allowlist, and can return "Enabled" directly. This way, the behavior will be the same with the [Define an inherited policy for feature in container at origin](https://www.w3.org/TR/permissions-policy/#algo-define-inherited-policy-in-container) algorithm with a "*" container policy.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -946,7 +946,7 @@ partial interface HTMLIFrameElement {
 
     <div class="algorithm" data-algorithm="is-feature-enabled">
     Given a feature (|feature|), a {{Document}} object
-    (|document|), an [=origin=] (|origin|), and a boolean isEquivalentOptInFeatureExistAndEnabled, this algorithm
+    (|document|), an [=origin=] (|origin|), and a boolean (|isEquivalentOptInFeatureExistAndEnabled|), this algorithm
     returns "<code>Disabled</code>" if |feature| should be considered
     disabled, and "<code>Enabled</code>" otherwise.</p>
     1. Let |policy| be |document|'s <a>Permissions Policy</a>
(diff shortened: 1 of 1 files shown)


jkarlin:
Can you modify this to support x-site javascript calls in addition to requests?

For instance, joinAdInterestGroup() allows you to set a x-site owner origin, so the interest group data will be written to the x-site's data. I believe this x-site owner should be checked in a similar way to your patch here for requests. That is, there should be some, `IsFeatureEnabledInAllowedSubframe(feature, origin)` type method that will verify if an `origin` iframe were created from this context with `allow=feature`, if the feature would be enabled. 

It seems like you could use this as for resource requests as well? That is, instead of adding all of the opted-in types, just add the opt-in type for the feature being checked.

xyaoinum:
@jkarlin :
This is meant to be a 'presumed' subframe, not a real subframe, right? This makes it a difficult concept to define. Using the term 'subframe' alone could create confusion, as there's already algorithm "Is feature enabled in document for origin?".

On the point of adding opt-in types, that was my initial thought (as seen in [this earlier commit](https://github.com/w3c/webappsec-permissions-policy/pull/499/commits/f3a53cef2dcc9d6e08c9d48e642eab782593bafb)), but I shifted to align with Ari's suggestion in the [code](https://chromium-review.googlesource.com/c/chromium/src/+/4111962/4..10/third_party/blink/public/common/permissions_policy/permissions_policy.h#b182): _"This feels a little dangerous as a boolean that's passed in might be better to pass in a set of opt_in features ..."_

@arichiv :
Ari, do you have thoughts on both the terminology for this 'presumed' subframe and the approach to handling opt-in types?

jkarlin:
I personally find the list to be confusing when we're only ever talking about checking a single feature at a time and prefer the boolean.

Looking again, I think your existing change (or ideally preferably with the boolean), is likely sufficient for my case, but it seems like it would be good to specify when the boolean should be set to true. You already have that text for requests in your proposal. How would we write it for javascript methods that are x-origin in nature?

Issue #480, opened 2022-07-19
Denying self while still allowing subframes


fergald:
Should it be possible to prevent a frame/origin from using a feature while still retaining the ability to let other, specified origins use it? 

This arose with the proposed `unload` feature (#444) but seems applicable to other features. There are 2 distinct uses
- gradual removal of a "bad" feature, e.g. `unload, `sync-xhr`. It may not be possible to remove all uses of the feature from 3rd-party subframes or that may just take time but a site may have removed all uses themselves and do not want new uses to be accidentally introduced. Currently there is no way to enforce this with permissions policy, we are forced to grant all ancestor-frames access to the feature
- reducing the attack surface for a "dangerous" feature. For example, limiting the use of some dangerous API to a different subdomain, e.g. example.com could place all uses of the `payment` feature in https://payment.example.com/ and deny use of that feature to `https://www.example.com/`. With this, compromising www.example.com does not provide access to this API and (hopefully) the JS on payment.example.com is smaller/simpler and harder to exploit.

Quote from @clelland in this [thread](https://groups.google.com/a/chromium.org/g/bfcache-dev/c/NoMgFMpJ1KU/m/nrBa1qcZAQAJ) on the history of this topic.

> I believe that this was the biggest reason -- with the original model, it was trivial to work around not having access yourself. Even with a header like "Feature-Policy: some-feature 'none'", if we made it possible to delegate without having access yourself, then you could just create a frame to anywhere, even your own origin, with an allow attribute that would re-enable the feature. The way we closed that hole that was to be strict about not allowing delegation when you don't have access yourself.

Now that the `allow` attribute is no longer enough by itself, the loophole does not exist any more and so preventing usage but allowing delegation makes sense.

# Syntax for discussion

For now, the syntax is not important but having a concrete proposal might be useful for discussing the idea.

One possible syntax would be to prepend `!` to indicate that the origin in question cannot use the feature, so for example

```
Permissions-Policy: payment=(!"https://www.example.com/" "https://payment.example.com/")
```

or

```
Permissions-Policy: payment=(!self "https://payment.example.com/")
```

annevk:
I think `not-self` or `exclude-self` is fine. I think the main problem is with `!` which in CSS means something completely different. We also don't really lack the space (header bloat issues notwithstanding).

fergald:
[...]

PR #501, opened 2022-12-20
Fix `Define an inherited policy for feature in container at origin`

arichiv:
The text implies only the inherited policy of the document should
be checked, but even if the containing document has the feature
enabled the document could set a permissions policy that denied access
to the specific origin. We must check the actual policy for the
containing document to know. Chrome is already doing this.

We can accomplish this by adapting both 9.8.2 and 9.8.3 to
use `Is feature enabled in document for origin?` but have the
first check the parent origin and the second check the
actual origin.

***
<a href="https://pr-preview.s3.amazonaws.com/arichiv/webappsec-permissions-policy/pull/501.html" title="Last updated on Dec 20, 2022, 3:04 PM UTC (5247072)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/501/b44d905...arichiv:5247072.html" title="Last updated on Dec 20, 2022, 3:04 PM UTC (5247072)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -920,13 +920,14 @@ partial interface HTMLIFrameElement {
     (|origin|), this algorithm returns the <a>inherited policy</a> for that
     feature.
     1. Let |policy| be |container|'s <a>node document</a>'s <a>Permissions
-      Policy</a>
+      Policy</a>.
+    1. If the result of executing <a abstract-op>Is feature
+      enabled in document for origin?</a> on |feature|, |container|'s <a>node document</a>',
+      and |container|'s <a>node document</a>''s |origin| is  "<code>Disabled</code>",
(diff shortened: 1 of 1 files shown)


arichiv:
Actually 9.8.3 covers this

clelland:
This looks good, and matches the intention, and, I think, the actual implementation in Chrome. Happy to merge this.

Note that this behaviour will change when https://github.com/w3c/webappsec-permissions-policy/issues/480 is fixed -- for that issue, we actually *do* want a way to block the use of a feature in the current frame, but retain the ability to delegate it to a subframe, but that will require parser changes, as well as support from this algorithm.

PR #505, opened 2023-01-11
Update for new navigable concepts.

clelland:
The browsing context container and nested browsing context concepts, used by the algorithms here, have been replaced in HTML with navigables and traversibles. Updates the spec to use those concepts instead.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/pull/505.html" title="Last updated on Jan 23, 2023, 5:30 AM UTC (99edaab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/505/f909827...99edaab.html" title="Last updated on Jan 23, 2023, 5:30 AM UTC (99edaab)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,6 +17,8 @@ spec:dom; type:interface; for:/; text:Document
 spec:url; type:dfn; for:url; text:origin
 spec:fetch; type:dfn; for:Response; text:response
 spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:dfn; for:/; text:navigable container
+spec:html; type:dfn; for:/; text:nested navigable
 spec:html; type:element; text:script
 spec:html; type:element; text:link
 spec:fetch; type:dfn; text:name
@@ -226,11 +228,11 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
     each feature, and whether it can be controlled by a <a>declared policy</a>
     in the document.
     </p>
-    <p>In a {{Document}} in a [=top-level browsing context=], the inherited
+    <p>In a {{Document}} in a [=/top-level traversable=], the inherited
     policy is based on defined defaults for each feature.</p>
-    <p>In a {{Document}} in a [=child browsing context=], the inherited policy
-    is based on the parent document's permissions policy, as well as the [=child
-    browsing context=]'s <a>container policy</a>.
+    <p>In a {{Document}} in a [=child navigable=], the inherited policy is based
+    on the parent document's permissions policy, as well as the [=child
+    navigable=]'s <a>container policy</a>.
     </div>
   </section>
   <section>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -17,6 +17,8 @@ spec:dom; type:interface; for:/; text:Document
 spec:url; type:dfn; for:url; text:origin
 spec:fetch; type:dfn; for:Response; text:response
 spec:html; type:dfn; for:/; text:browsing context
+spec:html; type:dfn; for:/; text:navigable container
+spec:html; type:dfn; for:/; text:nested navigable
```
domenic, line 21: These shouldn't be necessary, but I guess HTML failed to export these terms. Will fix!

You'll need to change nested navigable references to `[=navigable container/nested navigable=]`, when I do.

index.bs
```diff
@@ -248,12 +250,12 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
   </section>
   <section>
     <h3 id="container-policies">Container policies</h3>
-    <p>In addition to the <a>header policy</a>, each [=nested browsing context=]
-    has a <dfn export>container policy</dfn>, which is a <a>policy directive</a>,
+    <p>In addition to the <a>header policy</a>, each [=nested navigable=] has a
```
domenic, line 253: I think this and other references should be child navigable, not nested navigable. nested navigable is a property; child navigable is a data type.

index.bs
```diff
@@ -902,10 +904,10 @@ partial interface HTMLIFrameElement {
     Given a feature (|feature|), an <a>origin</a> (|origin|), and a <a>browsing
     context</a> (|browsingContext|), this algorithm returns the <a>inherited
     policy</a> for that feature.
-    1. If |browsingContext| is the [=nested browsing context=] of a [=browsing
-      context container=], return the result of executing <a abstract-op>Define
-      an inherited policy for feature in container at origin</a> for |feature|
-      in |browsingContext|'s <a>browsing context container</a> at |origin|.
+    1. If |browsingContext| is the [=nested navigable=] of a [=navigable
```
domenic, line 907: Here you're equating browsing contexts and navigables, which is not great. Ideally the algorithm should be updated to only operate on navigables.

index.bs
clelland: That's exactly the sort of thing I expected to mess up -- thanks!

domenic: Just opened https://github.com/whatwg/html/issues/8713 because this is too easy to mess up.

index.bs
clelland: Done!

index.bs
clelland: I've updated the interface with HTML now; we don't actually need browsing context to be passed in at all. I couldn't use navigable here previously, because these algorithms are called from https://html.spec.whatwg.org/#creating-a-new-browsing-context, before a navigable is created. However, we do have *embedder* at that point, which is what was actually being used.

*embedder* is currently just an Element; I'm pretty sure it's (currently) always going to be a navigable container, but I put an assertion in, so that if that ever changes, we know that this needs to be revisited.

index.bs
```diff
@@ -859,33 +858,36 @@ partial interface HTMLIFrameElement {
...
-          |origin| and |browsingContext|.
+          inherited policy for feature in container at origin</a> on |feature|,
+          |container| and |origin|.
         1. Set |inherited policy|[|feature|] to |isInherited|.
     1. Let |policy| be a new <a>permissions policy</a>, with inherited policy
       |inherited policy| and declared policy |declared policy|.
     1. Return |policy|.
 
+    Note: HTML should be updated to call this algorithm from "create a new
+    browsing context and document".
```
domenic, line 879: This shouldn't be necessary, hopefully, since we'll just merge the two PRs in close proximity.

domenic (approved):
Maybe also update

> SecureCorp Inc. wants to completely disable use of the Geolocation API within all browsing contexts except for its own origin

? I think maybe [=Document/descendant navigables=] captures what you mean here?

index.bs
clelland: Removed!

PR #511, opened 2023-05-07
Link to more rigorous definitions

### LINKED ISSUE #489: "Process permissions policy attributes" unclear
This algorithm should only work for `iframe` elements, no? But step 1 seems awfully generic.

It also doesn't address the case when there is no `allow` attribute or element is not an `iframe` element.

domfarolino:
This PR cleans up various parts of the Permissions Policy standard by:
 - Formalizing links to various definitions such as `[=Document/origin=]`, plenty of https://infra.spec.whatwg.org/ concepts, and internal definitions that we're linked to. This is perhaps most of this PR.
 - Gives to formal members the permissions policy struct, namely the inherited policy and declared policy, whose "type" are the definitions of the same name that are not `for=""` anything. This makes it much more clear which every instance of an IP or DP belong to, etc.

This also fixes #489.

***
<a href="https://pr-preview.s3.amazonaws.com/domfarolino/webappsec-permissions-policy/pull/511.html" title="Last updated on May 23, 2023, 8:29 PM UTC (40aae52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/511/b363be8...domfarolino:40aae52.html" title="Last updated on May 23, 2023, 8:29 PM UTC (40aae52)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -321,12 +321,12 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
       <dd>The feature is allowed in [=navigable/active document|documents=] in
       [=/top-level traversables=] by default, as well as those in [=child
       navigables=] whose [=navigable/active document|document=] is
-      [=same origin=] with the [=/top-level traversable=]'s [=navigable/active
-      document|document=]. It is disallowed by default in [=child navigables=]
-      whose [=navigable/active document|document=] is cross-origin with its
-      [=navigable/parent=]'s [=navigable/active document|document=], but can be
-      enabled by explicitly supplying a [=container policy=] that delegates the
-      feature.</dd>
+      [=same origin=] with its [=navigable/parent=]'s [=navigable/active
+      document|document=], when allowed in that {{Document}}. It is disallowed
+      by default in [=child navigables=] whose [=navigable/active
+      document|document=] is cross-origin with its [=navigable/parent=]'s
+      [=navigable/active document|document=], but can be enabled by explicitly
+      supplying a [=container policy=] that delegates the feature.</dd>
     </dl>
   </section>
 </section>
--- a/index.bs
+++ b/index.bs
@@ -201,10 +201,10 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
 
     <dl dfn-for="permissions policy">
       : <dfn>inherited policy</dfn>
-      :: an <a for=/>inherited policy</a>
+      :: an [=ordered map=] from [=features=] to "`Enabled`" or "`Disabled`"
 
       : <dfn>declared policy</dfn>
-      :: an <a for=/>declared policy</a>
+      :: an [=ordered map=] from [=features=] to [=allowlists=]
     </dl>
     <p>An <dfn export>empty permissions policy</dfn> is a <a>permissions
     policy</a> that has an <a for="permissions policy">inherited policy</a> which
@@ -213,10 +213,6 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
   </section>
   <section>
     <h3 id="inherited-policies">Inherited policies</h3>
-    <p>An <dfn data-lt="inherited policy|inherited policies">inherited
-    policy</dfn> is an [=ordered map=] from <a
-    data-lt="policy-controlled feature">features</a> to either
-    "<code>Enabled</code>" or "<code>Disabled</code>".</p>
     <p>The <dfn export>inherited policy for a feature</dfn> <var>feature</var>
     is the value in the <a for="permissions policy">inherited policy</a> whose
     key is <var>feature</var>. After a <a>permissions policy</a> has been
@@ -236,13 +232,6 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
     on the parent document's permissions policy, as well as the [=child
     navigable=]'s <a>container policy</a>.
     </div>
-  </section>
-  <section>
-    <h3 id="declared-policies">Declared policies</h3>
-    <p>A <dfn data-lt="declared policy|declared permissions policy">declared
-    policy</dfn> is an [=ordered map=] from
-    <a data-lt="policy-controlled feature">features</a> to <a>allowlists</a>.
-    </p>
   </section>
   <section>
     <h3 id="header-policies">Header policies</h3>
@@ -759,7 +748,7 @@ partial interface HTMLIFrameElement {
 
     <div class="algorithm" data-algorithm="process-response-policy">
     Given a [=response=] (|response|) and an [=origin=] (|origin|), this
-    algorithm returns a <a>declared permissions policy</a>.
+    algorithm returns a <a for="permissions policy">declared policy</a>.
 
     1. Let |parsed header| be the result of executing <a>get a structured
       field value</a> given "<code>Permissions-Policy</code>" and "dictionary" from
@@ -776,7 +765,7 @@ partial interface HTMLIFrameElement {
 
     <div class="algorithm" data-algorithm="construct-policy">
     Given an <a>ordered map</a> (|dictionary|) and an [=origin=] (|origin|), this
-    algorithm will return a <a>declared permissions policy</a>.
+    algorithm will return a <a for="permissions policy">declared policy</a>.
     1. Let |policy| be an empty [=ordered map=].
     1. [=map/For each=] |feature-name| → |value| of |dictionary|:
         1. If |feature-name| does not identify any recognized
--- a/index.bs
+++ b/index.bs
@@ -294,7 +294,7 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
       stored in the allowlist. The keywords themselves are not part of the
       allowlist.
     </div>
-    <div algorithm=matches>
+    <div algorithm="matches">
       <p>To determine whether an <a>allowlist</a> <dfn>matches</dfn> an origin
       <var>origin</var>, run these steps:
 
--- a/index.bs
+++ b/index.bs
@@ -14,12 +14,14 @@ Markup Shorthands: css no, markdown yes
 </pre>
 <pre class="link-defaults">
 spec:dom; type:interface; for:/; text:Document
+spec:dom; type:dfn; for:/; text:element
 spec:url; type:dfn; for:url; text:origin
 spec:fetch; type:dfn; for:Response; text:response
 spec:html; type:element; text:script
 spec:html; type:element; text:link
 spec:fetch; type:dfn; text:name
 spec:fetch; type:dfn; text:value
+spec:infra; type:dfn; text:list
 </pre>
 <pre class="anchors">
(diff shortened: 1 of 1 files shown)


clelland (commented):
Thanks for this cleanup! It's long overdue, I think. I've added some notes inline, but this is really great.

domfarolino:
Oops, I left this as a draft since I had a few questions/nits to make but am mostly OOO-ish this week. I'll address the above soon and maybe put some more comments myself.

index.bs
```diff
@@ -290,41 +296,49 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      then return true.</li>
-      <li>Otherwise, for each <var>item</var> in the <a>allowlist</a>:
-        <ol>
-          <li>If <var>item</var> is [=same origin-domain=] with
-          <var>origin</var>, then return true.</li>
-        </ol>
-      </li>
-      <li>return false.</li>
-    </ol>
+    <div algorithm=matches>
```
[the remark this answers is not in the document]
domfarolino: Up to you. Provided an attribute value does not have spaces, there is no requirement for quotes, and some specs never use them ([here is an example](https://github.com/WICG/fenced-frame/blob/master/spec.bs#L313-L316) of not using them in several attributes, e.g.,`for=` and `algorithm=`).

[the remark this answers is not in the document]
clelland: Yeah, I guess I knew they were optional :)
I'll just subscribe to the "match surrounding style and hope for eventual consistency" school :)

index.bs
```diff
@@ -900,10 +917,10 @@ partial interface HTMLIFrameElement {
 
     <div class="algorithm"
     data-algorithm="define-inherited-policy-in-container">
-    Given a feature (|feature|), null or a <a>navigable container</a>
-    (|container|), and an <a>origin</a> for a document in that container
-    (|origin|), this algorithm returns the <a>inherited policy</a> for that
-    feature.
+    Given a [=feature=] (|feature|), null or a <a>navigable container</a>
+    (|container|), and an <a for="Document">origin</a> for a {{Document}} in
+    that container (|origin|), this algorithm returns the <a for=/>inherited
```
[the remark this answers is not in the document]
domfarolino: Yeah I think this goes to the larger question of whether or not we need a `for=/` dfn for "inherited policy" in the first place anyways, now that this PR gives the `for="permissions policy"` definition of "inherited policy".

But yeah for this specific comment, it's probably not right either way. I think this should use "inherited policy for a feature" since that's really what we're referencing. I'll change this.

index.bs
```diff
@@ -290,41 +296,49 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      default, and when allowed, is allowed by default to same-origin domain
-      documents in child navigables, but is disallowed by default in
-      cross-origin documents in child navigables.</dd>
+      <dd>The feature is allowed in [=navigable/active document|documents=] in
+      [=/top-level traversables=] by default, as well as those in [=child
+      navigables=] whose [=navigable/active document|document=] is
+      [=same origin=] with the [=/top-level traversable=]'s [=navigable/active
+      document|document=]. It is disallowed by default in [=child navigables=]
+      whose [=navigable/active document|document=] is cross-origin with the
+      [=/top-level traversable=]'s [=navigable/active document|document=], but
```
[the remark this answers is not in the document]
domfarolino: Is this true? It can't always be enabled in a child navigable via the `Permisisons-Policy` header inside. The only thing that header could do *for sure* (without being dependent on the container policy) is _disable_ the feature, which I've clarified above. WDYT?

domfarolino:
Hey Ian, please look specifically at the last commit where I actually remove the top-level definitions of "inherited policy" and "declared policy" altogether, in favor of the definitions that are `for="permissions policy"` specifically. This resulted in removing the [Declared policies](https://w3c.github.io/webappsec-permissions-policy/#declared-policies) section altogether, however I left the inherited one alone, since it exports the definition https://w3c.github.io/webappsec-permissions-policy/#inherited-policy-for-a-feature that is _not_ redundant with the `for="permissions-policy"` definitions.

domfarolino:
Ping @clelland

index.bs
```diff
@@ -195,79 +197,70 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+      :: an [=ordered map=] from [=features=] to "`Enabled`" or "`Disabled`"
+
+      : <dfn>declared policy</dfn>
+      :: an [=ordered map=] from [=features=] to [=allowlists=]
+    </dl>
     <p>An <dfn export>empty permissions policy</dfn> is a <a>permissions
-    policy</a> that has an <a>inherited policy</a> which contains
-    "<code>Enabled</code>" for every <a>supported feature</a>, and a <a>declared
+    policy</a> that has an <a for="permissions policy">inherited policy</a> which
+    contains "<code>Enabled</code>" for every <a>supported feature</a>, and a <a for="permissions policy">declared
```
clelland, line 211: Nit: wrapping

index.bs
```diff
@@ -290,41 +283,50 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
+      suitable <a http-header>`Permissions-Policy`</a> header).</dd>
       <dt><dfn for="default allowlist" export><code>'self'</code></dfn></dt>
-      <dd>The feature is allowed in documents in top-level traversables by
-      default, and when allowed, is allowed by default to same-origin domain
-      documents in child navigables, but is disallowed by default in
-      cross-origin documents in child navigables.</dd>
+      <dd>The feature is allowed in [=navigable/active document|documents=] in
+      [=/top-level traversables=] by default, as well as those in [=child
+      navigables=] whose [=navigable/active document|document=] is
+      [=same origin=] with the [=/top-level traversable=]'s [=navigable/active
```
clelland, line 324: I think this should still read something like "`...is [=same origin=] with its [=navigable/parent=]'s [=navigable/active document|document=], when allowed in that {{Document}}`"

index.bs
```diff
@@ -644,15 +647,15 @@ partial interface HTMLIFrameElement {
     |document|'s permissions policy.</p>
     <p>To get the <a>observable policy</a> for an Element |node|, run the
     following steps:</p>
-        1. Let |inherited policy| be a new ordered map.
-        2. Let |declared policy| be a new ordered map.
-        3. For each <a>supported feature</a> |feature|:
+        1. Let |inherited policy| be a new [=ordered map=].
```
clelland, line 650: The inconsistency is mine, originally, but this algorithm uses "a new ordered map", while others use "an empty ordered map" to mean the same thing. Should we commit to one or the other?

(Or feel free to ignore; this is a pre-existing issue, and you don't have to clean up *everything* :) )

clelland: (I'm not yet sure how I feel about the syntax, but we can also use `«[ ]»` as literal syntax for an empty map.)

index.bs
```diff
@@ -841,18 +846,17 @@ partial interface HTMLIFrameElement {
     <div class="algorithm" data-algorithm="process-policy-attributes">
     Given an element (|element|), this algorithm returns a <a>container
     policy</a>, which may be empty.
+    1. If |element| is not an <{iframe}> element, then return an empty [=policy
```
clelland, line 849: Do you want to mention in the commit message that this fixes #489 ?

index.bs
domfarolino: No strong opinions myself. I've updated this to use the `an empty [=ordered map=]` syntax you recommend for consistency.

index.bs
domfarolino: Aren't both true? For example, my understanding is that in `A(B(A))` if the feature is disabled in `B`, the inner `A` still has it enabled if the outer `A` does, since the inner `A` could just synchronously script its grandparent and use the feature anyways (so there's no need to turn it off for the inner-most `A` here).

I think your statement is also true as well, but maybe we could outline both scenarios here?

domfarolino:
OK it looks like there are only two outstanding comments at this point, which could use your input:
 - https://github.com/w3c/webappsec-permissions-policy/pull/511#discussion_r1189042015
 - https://github.com/w3c/webappsec-permissions-policy/pull/511#discussion_r1199294407

index.bs
clelland: In `A(B(A))`, if it's disabled in `B`, it will be disabled in inner `A` as well -- it's true that inner `A` could usually call back to outer `A` to use a feature (depending on the feature; some would act differently if called in different frames), but to make the algorithms consistent, we disable it -- generally keeping to "if a feature is disabled in any frame, then subframes cannot re-enable it".

So there's no same-origin-with-top-level check (or really a same-origin-with-any-other-frame-with-the-feature-enabled check); we just check whether each frame embedding creates an origin boundary, and determine inheritance at that point.

index.bs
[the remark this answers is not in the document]
clelland: > disallowed by default in child navigables whose document is cross-origin with the top-level traversable's document

The point here is that we don't check, for a given non-top-level frame, whether it's cross-origin with the top-level; we check its relationship with it's parent frame. If it's cross-origin with it's parent frame, then a default-self feature will be disabled in the child unless the parent frame `allow`s it. (It's not a sufficient condition; there are other conditions that need to be met as well, like the feature actually being available to the parent frame, and the child frame not declining the feature through a `permissions-policy` header). Regardless of whether it's same-origin with the top-level frame, default-self features are disabled across origin boundaries by default.

index.bs
domfarolino: Got it. I've clarified this here.

index.bs
```diff
@@ -290,41 +283,50 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
...
-      cross-origin documents in child navigables.</dd>
+      <dd>The feature is allowed in [=navigable/active document|documents=] in
+      [=/top-level traversables=] by default, as well as those in [=child
+      navigables=] whose [=navigable/active document|document=] is
+      [=same origin=] with its [=navigable/parent=]'s [=navigable/active
+      document|document=], when allowed in that {{Document}}. It is disallowed
+      by default in [=child navigables=] whose [=navigable/active
+      document|document=] is cross-origin with its [=navigable/parent=]'s
+      [=navigable/active document|document=], but can be enabled by explicitly
+      supplying a [=container policy=] that delegates the feature.</dd>
```
domfarolino, line 329: @clelland Given some of the discussion in https://github.com/w3c/webappsec-permissions-policy/pull/511#discussion_r1187708674 and https://github.com/w3c/webappsec-permissions-policy/pull/511#discussion_r1199294407, let me know if you think I should just remove the sentence from lines 325-329, since technically other conditions have to be satisfied _besides_ a lenient container policy.

domfarolino:
OK I think this should just about be ready % one question about a comment.

index.bs
clelland: Maybe delete everything from "but can be", since the previous part of the statement is true. The bit about "can be enabled with a container policy" is the part that might also require other conditions. Does that make sense?

clelland (approved):

index.bs
domfarolino: Yep, definitely. The conditions seem too many to enumerate here, so paring this down makes sense. Done.

PR #515, opened 2023-05-26
add none as an allowed default

### LINKED ISSUE #513: allow disabled-by-default features
# Motivation:

This would be a tool to use in deprecating and removing a feature. Our current motivator for this is the [deprecation of unload](https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md) but it could apply to any other feature.

Deprecating is hard and we want to provide a period where a feature is still usable but opt-in. The [plan for `unload`](https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md) is more complicated than just flipping to this state but this is the desired state.

Currently the [available defaults](https://www.w3.org/TR/permissions-policy/#default-allowlists) for a feature are `self` and `*`. It is not possible to have a policy that disables a feature by default.

# Proposal

Add an allowed default value of `none`. This would mean that by default, the feature is not available. To make the feature available in a given frame, that frame would have to declare it via the header and every ancestor of that frame would also have to declare it via the header and delegate it via the `allow` attribute on the `iframe` element.

This

- fits with the existing PP model
- achieve goal of allowing opt-back-in
- opt-in requires action on by allow frames in the ancestry chain

The unfortunate side-effect of forcing the feature to be enabled in a the ancestor frames even if they do not want it could be avoided by also implementing [`not-self` ](https://github.com/w3c/webappsec-permissions-policy/issues/480).

# Alternative outcomes

There are some different outcomes one could imaging trying to achieve.

## require action only by the subframe itself

- by subframes can enable feature
- ancestors can epxlicitly prevent this

This would make it easier to opt back in while still giving containers utlimate control. It does not seem to sit with the existing inheritance model of PP and would require changes to the decision algorithm, not just the available defaults.

## require action only by the containing frames

- subframes don't need to explicitly enable the feature
- ancestors enable the feature

This would allow sites to restore `unload` functionality to 3rd party frames even without the cooperation of the 3rd party. It seems like 3rd parties have an incentive to enable the features they use, so this is of questionable benefit vs the proposed outcome.

@clelland

fergald:
Add `none` as a new possible default allowlist.

Surprisingly, I don't think this requires any updates to the algorithms as the updates would be to "Define an inherited policy for feature in container at origin" and "Is feature enabled in document for origin?" would both just be adding "If feature’s default allowlist is None, return "Disabled"." where we are going to return "Disabled" anyway.

Maybe we should add these conditions anyway, to make it explicit. Up to you.

Fixes #513

@clelland

***
<a href="https://pr-preview.s3.amazonaws.com/fergald/webappsec-permissions-policy/pull/515.html" title="Last updated on Jun 21, 2024, 6:22 AM UTC (59b2c00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/515/7a2bc78...fergald:59b2c00.html" title="Last updated on Jun 21, 2024, 6:22 AM UTC (59b2c00)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -399,7 +399,7 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
       navigables=].  It can be allowed {{Document}}s in [=/top-level
       traversables=] by delivering the {{Document}} with a suitable <a
       http-header> `Permissions-Policy`</a> header). It can be allowed
-      in in [=child navigables=] if it allowed in the parent
+      in [=child navigables=] if it is allowed in the parent
       {{Document}} by explicitly supplying a [=container policy=] on
       the [=navigable container=] that overrides this default and by
       delivering the {{Document}} with a suitable <a
--- a/index.bs
+++ b/index.bs
@@ -1056,7 +1056,9 @@ partial interface HTMLIFrameElement {
            policy</a>'s [=declared policy/declarations=][|feature|]
            <a>matches</a> |origin|, then return "<code>Enabled</code>".
         1. Otherwise return "<code>Disabled</code>".
-    1. Return "<code>Enabled</code>".
+    1. If |feature|'s <a>default allowlist</a> is <code>*</code> or
+       <code>self</code>, return "<code>Enabled</code>".
+    1. Return "<code>Disabled</code>".
 
     </div>
   </section>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1068,9 +1081,11 @@ partial interface HTMLIFrameElement {
         1. Otherwise return "<code>Disabled</code>".
     1. If |feature|'s <a>default allowlist</a> is <code>*</code>, return
       "<code>Enabled</code>".
+    1. Let |document origin| be |document|'s [=Document/origin=].
     1. If |feature|'s <a>default allowlist</a> is <code>'self'</code>, and
       |origin| is [=same origin=] with |document origin|, return
       "<code>Enabled</code>".
+    1. If |document| is a headerless document, return "<code>Enabled</code>".
```
clelland, line 1088: What about defining (somewhere else in the doc):

```
A document is a <dfn>headerless document<dfn> if any of the following conditions are true:
* Its [=Document/URL=] [=matches about:blank=]
* Its [=Document/URL=] [=matches about:srcdoc=]
* Its [=Document/URL=]'s [=scheme=] is 'data'
```

And we can add additional cases as we find them. This might also be a good candidate for export to HTML if it turns out to be generally useful.

Solomonus (approved):

bobomb:
@fergald Due to https://github.com/w3c/webauthn/issues/2430 I'd like to see this move forward so it can be implemented fully in Chromium, and would greatly appreciate if you can help move this along? And if you need anything from me to help move it along, I would gladly step in to get it going.

fergald:
@bobomb I'm happy to rebase this and get it committed if there is agreement. It looks like currently there's some argument against a default of `none`. If you can get that cleared up and point to agreement from another browser vendor, I expect @clelland would accept this PR.

PR #516, opened 2023-06-01
Wildcards in Permissions Policy Origins

### LINKED ISSUE #479: Client Hint delegation to multiple subdomains

arichiv:
This change harnesses [Content Security Policy matching](https://www.w3.org/TR/CSP3/#matching-urls) to allow wildcards in the origins of Permissions Policy allowlists. The goal of this is to provide greater granularity in permissions to better balance the ease and risk of a wildcard permission against what might otherwise be a long list of similar, yet distinct origins.

The only wildcard allowed before was *, which matched all origins.

Wildcards will now be supported for:
* https://example.com:\* (ports, any port for example.com over https)
* https://\*.com (hosts, any https origin ending in .com)
* https: (scheme-only, any https origin)

Feature: https://chromestatus.com/feature/5101218029895680

closes #479

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/pull/516.html" title="Last updated on Jun 9, 2023, 2:56 PM UTC (ae9d1b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/516/388f33c...ae9d1b6.html" title="Last updated on Jun 9, 2023, 2:56 PM UTC (ae9d1b6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -38,6 +38,7 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
   type: dfn
     text: scheme-source; url: grammardef-scheme-source
     text: host-source; url: grammardef-host-source
+    text: Does url match expression in origin with redirect count?; url: match-url-to-source-expression
 </pre>
 <pre class="biblio">
 {
@@ -305,7 +306,7 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
     <h3 id="policy-directives">Policy directives</h3>
     <p>A <dfn data-lt="policy directive|policy directives">policy
     directive</dfn> is an [=ordered map=], mapping <a>policy-controlled
-    features</a> to corresponding <a>allowlists</a> of origins.</p>
+    features</a> to corresponding <a>allowlists</a> of <a>permissions-source-expression</a>s.</p>
     <p>A <a>policy directive</a> is represented in HTTP headers as the
     serialization of an <a>sh-dictionary</a> structure, and in and HTML
     attributes as its ASCII serialization.</p>
@@ -314,11 +315,11 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
     <h3 id="allowlists">Allowlists</h3>
     <p>A permissions policy <dfn export
     lt="allowlist|allowlists">allowlist</dfn> is conceptually a set of
-    [=origins=]. An <a>allowlist</a> may be either:
+    <a>permissions-source-expression</a>s. An <a>allowlist</a> may be either:
     <ul>
       <li><dfn>The special value <code>*</code></dfn>, which represents every
       origin, or</li>
-      <li>An <a>ordered set</a> of <a>scheme-source</a>s and <a>host-source</a>s</li>
+      <li>An <a>ordered set</a> of <a>permissions-source-expression</a>s</li>
     </ul>
     <div class="note">
       The keywords <code>'self'</code>, <code>'src'</code>, and
@@ -337,8 +338,8 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
 
       1. Otherwise, [=set/for each=] |item| in the <a>allowlist</a>:
 
-        1. If |item| is [=same origin-domain=] with <var>origin</var>, then
-           return true.
+        1. If the result of running <a>Does url match expression in origin with redirect count?</a>
+           on <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
 
       1. Return false.
     </div>
@@ -385,7 +386,8 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
       <dfn>serialized-policy-directive</dfn> = <a>feature-identifier</a> RWS <a>allow-list</a>
       <dfn>feature-identifier</dfn> = 1*( ALPHA / DIGIT / "-")
       <dfn>allow-list</dfn> = <a>allow-list-value</a> *(RWS <a>allow-list-value</a>)
-      <dfn>allow-list-value</dfn> = <a>scheme-source</a> / <a>host-source</a> / "*" / "'self'" / "'src'" / "'none'"
+      <dfn>allow-list-value</dfn> = <a>permissions-source-expression</a> / "*" / "'self'" / "'src'" / "'none'"
+      <dfn>permissions-source-expression</dfn> = <a>scheme-source</a> / <a>host-source</a>
     </pre>
     <div class="note">
       The string "<code>'self'</code>" may be used as an origin in an
@@ -407,7 +409,7 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
     feature, then the Dictionary Member will be ignored by the processing steps.
 
     The Member Values represent <a>allowlists</a>, and must be one of:
-    * a String containing the ASCII <a>serialization of an origin</a>
+    * a String containing the ASCII <a>permissions-source-expression</a>
     * the Token `*`
     * the Token `self`
     * an Inner List containing zero or more of the above items.
@@ -673,9 +675,8 @@ partial interface HTMLIFrameElement {
     4. If |feature| is not allowed in |policy| for |origin|, return |result|
     5. Let |allowlist| be |policy|'s declared policy[|feature|]
     6. If |allowlist| is the special value `*`, append "`*`" to |result|
-    7. Otherwise, for each |origin| in |allowlist|:
-        1. Append the <a lt="serialization of an origin">serialization</a> of
-            |origin| to |result|
+    7. Otherwise, for each |item| in |allowlist|:
+        1. Append |item| to |result|
     8. Return |result|.
 
     <p>The <dfn>observable policy</dfn> for any Node is a <a>permissions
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -269,11 +315,11 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
     <h3 id="allowlists">Allowlists</h3>
     <p>A permissions policy <dfn export
     lt="allowlist|allowlists">allowlist</dfn> is conceptually a set of
-    [=origins=]. An <a>allowlist</a> may be either:
+    <a>permissions-source-expression</a>s. An <a>allowlist</a> may be either:
```
clelland, line 318: I would still maintain that it is a set of origins, conceptually. It's all of the origins (potentially infinite) which match the various patterns used in the header or attribute. The match algorithm is essentially testing set inclusion.

Structurally, it is definitely a set of expressions, as you mention below. (This is all pretty pedantic, though)

index.bs
```diff
@@ -292,8 +338,8 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
 
       1. Otherwise, [=set/for each=] |item| in the <a>allowlist</a>:
 
-        1. If |item| is [=same origin-domain=] with <var>origin</var>, then
-           return true.
+        1. If the result of running <a>Does url match expression in origin with redirect count?</a>
+           on <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
```
clelland, line 342: The 0 here is irrelevant, right? It looks like that only comes into effect if there is a path component, which we won't pass in?

clelland: Also, I'm not sure that we can pass `origin` in as the first parameter; the algorithm takes a URL, which an origin is not.

The second origin is interesting -- it should actually be the origin of the policy, which isn't part of the allowlist, since the algorithm handles items like 'self' here. However, we explicitly expand 'self' below when constructing allowlists. I wonder if this should be some null value instead, so that it cannot affect the match in any way?

index.bs
```diff
@@ -292,8 +338,8 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
```
clelland, line 338: We might want to revisit whether this step (above) is still necessary, and if so, point out that the behaviour diverges from CSP here.

arichiv: Added a note

index.bs
arichiv: We need to pass in the 0 since the algo asks for it though right?

Should I convert the origin to a string and then parse it as a URL? I assumed we could pass it here since it has a subset of the parts.

In the chrome source we are just passing the origin of the page twice https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/permissions_policy/origin_with_possible_wildcards.cc;drc=59515e08807151f06afcacfb2063732dd403ea05;l=116. I think that's the correct thing to do in this instance since we're doing all checks relative to the origin of the prospective page.

clelland: > We need to pass in the 0 since the algo asks for it though right?

Yeah, I just wanted to make sure that it didn't actually matter -- you could pass in 7, or 100000000, and it wouldn't make a difference since there should be no path component.

> Should I convert the origin to a string and then parse it as a URL? I assumed we could pass it here since it has a subset of the parts.

`origin` can be opaque, in which case there would be no valid serialization, and we actually do need to see if it is same-origin with `item`. At least, that's the case currently. I think there may be no way for an allowlist to contain an opaque origin with this change.

That might be a more subtle change in the spec than I thought at first: Initially an allowlist would only contain concrete origins which were all resolved during parsing (unless the whole allowlist was `*`), but now it contains a list of patterns to match against, which cannot be made concrete.

> In the chrome source we are just passing the origin of the page twice https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/permissions_policy/origin_with_possible_wildcards.cc;drc=59515e08807151f06afcacfb2063732dd403ea05;l=116. I think that's the correct thing to do in this instance since we're doing all checks relative to the origin of the prospective page.

That algorithm, as I read it, takes \<absolute url of resource to check\>, \<pattern to check against\>, \<origin to resolve pattern against\>, \<redirect count\>. Because \<pattern\> could be `self`, which needs to resolve to the containing page's URL, that second origin should be the policy-owning-page's origin.

I think it doesn't matter right now, because we pre-resolve items like `self` during parsing, and don't store them in the allowlist, but if we need to change that, then the second origin will need to be that of the page with the policy.

arichiv: Would it be enough to add a check that the origin wasn't opaque before calling this then?

index.bs
```diff
@@ -821,19 +872,22 @@ partial interface HTMLIFrameElement {
           |allowlist| to <a>the special value <code>*</code></a>.
         1. Otherwise:
             1. If |targetlist| is empty and |target origin| is given,
-               [=list/append=] |target origin| to |allowlist|.
+               let the [=origin=] representing `self` in |allowlist| be |target origin|.
```
clelland, line 875: This should be `src`, I believe.

index.bs
```diff
@@ -821,19 +872,22 @@ partial interface HTMLIFrameElement {
...
             1. For each |element| in |targetlist|:
                 1. If |element| is an <a>ASCII case-insensitive</a> match for
-                  "<code>'self'</code>", let |result| be |container origin|.
+                  "<code>'self'</code>":
+                    1. Let the [=origin=] representing `self` in |allowlist| be |target origin|.
```
clelland, line 879: Self should add `|container origin|` to the allowlist

clelland (commented):
I think this is looking pretty good --  a couple of issues with `allow` parsing though.

clelland:
Thanks -- I think this looks good -- can you file an issue on CSP to export the three terms that we use here? (That shouldn't block merging this, but it would be good cleanup for later)

The only thing that we might want to do to tighten it up a bit would be to give Allowlist two actual properties, so that we don't have to keep referring to "the origin representing `self` in allowlist", and could actually link to a definition there. That would probably require accounting for those properties to possibly be null. We'd be able to assume at that point that the "list" part of the allowlist is *just* patterns, as well.

(Not necessary right now; this is functionally equivalent, but if you wanted to do that before merging, let me know and I'll hold off)

arichiv:
Filed https://github.com/w3c/webappsec-csp/issues/604

I think this can merge as is, I can circle back on that cleanup when the DFNs are exported by CSP.

clelland (approved):

Issue #512, opened 2023-05-08
Potential bug in access delegation in cross-origin iframe for feature that has default allowlist value "self"?


shuranhuang:
According to the explainer, when no declared policy presents, a powerful features that has default allowlist value "`self`" should be allowed in the top-level document and its same-origin frames, but blocked in cross-origin frames, unless `allow` attribute is used to set this policy and the origin of the document in the frame matches the iframe's `src` attribute.

However, the following step from the "[Define an inherited policy for feature in container at origin](https://www.w3.org/TR/permissions-policy/#algo-define-inherited-policy-in-container)" algorithm, 
https://github.com/w3c/webappsec-permissions-policy/blob/b363be8b113ed980b0449945483920bf1a19bac3/index.bs#L912-L915
which steps into "[Is feature enabled in document for origin?](https://www.w3.org/TR/permissions-policy/#algo-is-feature-enabled)" algorithm to the following steps
https://github.com/w3c/webappsec-permissions-policy/blob/b363be8b113ed980b0449945483920bf1a19bac3/index.bs#L949-L952
will return "Disabled" for the case mentioned at the beginning, where "Enabled" is expected.

Could this be a bug or I am missing something here?

clelland:
That does seem like a bug -- that step should be checking to see if the feature can potentially be delegated to the child frame, or if it is blocked by the policy in the containing document. Either 9.8 needs to return `Enabled` in all default cases, or 9.7 should be calling something different in this step.

This algorithm needs to be rewritten to handle #480 as well, so I'll see if I can come up with text that fits that *and* matches the intended behavior for default-self inheritance.

shuranhuang:
Thanks for confirming! I think this issue was first introduced in https://github.com/w3c/webappsec-permissions-policy/pull/501. The fix should be easy. But I will leave it to you if you prefer to address this issue together with other issues.

PR #517, opened 2023-06-01
Fix "Define an inherited policy for feature in container at origin" algo

shuranhuang:
This fixes the issue https://github.com/w3c/webappsec-permissions-policy/issues/512 by replacing the algorithm used by `Define an inherited policy for feature in container at origin` for step#2 and step#3 from `Is feature enabled in document for origin?` to a new algorithm, which only checks whether the feature is allowed for origin with the parent frame's inherited policy and declared policy. Chrome is already doing this.

Note that we cannot let `Is feature enabled in document for origin?` to adopt this new algorithm because `Is feature enabled in document for origin?` needs to return `true` early when the feature is allowed by both inherited policy and declared policy, whereas the new algorithm could also `true` for when the feature is absent from the declared policy but allowed by inherited policy.

Fixes https://github.com/w3c/webappsec-permissions-policy/issues/512

***
<a href="https://pr-preview.s3.amazonaws.com/shuranhuang/webappsec-permissions-policy/pull/517.html" title="Last updated on Jun 5, 2023, 5:56 PM UTC (3f55b63)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/517/388f33c...shuranhuang:3f55b63.html" title="Last updated on Jun 5, 2023, 5:56 PM UTC (3f55b63)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -909,14 +909,13 @@ partial interface HTMLIFrameElement {
     that container (|origin|), this algorithm returns the [=inherited policy for
     a feature|inherited policy value=] for |feature|.
     1. If |container| is null, return "<code>Enabled</code>".
-    1. If the result of executing <a abstract-op>Is feature
-      enabled in document for origin?</a> on |feature|, |container|'s
-      <a>node document</a>, and |container|'s <a>node document</a>'s origin
-      is "<code>Disabled</code>", return "<code>Disabled</code>".
-    1. If the result of executing <a abstract-op>Is feature
-      enabled in document for origin?</a> on |feature|, |container|'s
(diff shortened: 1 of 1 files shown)


clelland:
This seems correct, and fixes the issue that was introduced with #378 where now the same-origin check should just be with the `allow` attribute when constructing the container policy. The header can be used to explicitly deny delegation to an origin, but by default that delegation is permitted.

default-self features still require that the origin be in the allowlist (if cross-origin with their container) but that happens in the remaining steps of this algorithm.

Issue #519, opened 2023-06-16
Does url match expression in origin with redirect count? takes a URL, not an origin


annevk:
https://w3c.github.io/webappsec-permissions-policy/#allowlists doesn't appear to invoke it correctly. Origins aren't compatible with URLs.

clelland:
This is probably going to have to end up using something like "let |url| be the result of calling the [[=url parser=]](https://url.spec.whatwg.org/#concept-url-parser) on the [[=serialization of an origin|serialization=]](https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin) of |origin|" to get an actual URL that can be used.

annevk:
Yeah, though perhaps we could redo part of CSP whereby the first half of the algorithm or so becomes its own algorithm that we'd pass url's origin and Permissions Policy can invoke directly.

How does it actually work for paths today?

clelland:
Do you mean paths in Permissions Policy declarations? I believe that in practice, paths have always been dropped from the string, and only the origin is used. (Chromium parses the URL, and then extracts the origin, throwing away the userinfo/path/query/fragment/whatever-else)

annevk:
Hmm, it doesn't really make sense to me that you'd extract an origin from an expression. Perhaps you mean an origin-expression from a url-expression or some such? I can see ignoring/dropping the path though in the expression.

clelland:
Sure -- before we had wildcard support, we would parse the string as a URL (assuming it wasn't `self`, `src` or `*`), and store only the origin of the parsed URL.

Now, I believe that we invoke the CSP parser on that string instead, which returns a [scheme/host/port/path struct](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:services/network/public/mojom/content_security_policy.mojom;l=53), and we clear the path member before storing it in the allowlist.

PR #523, opened 2023-06-28
Use exported DFNs from CSP

arichiv:
Added in https://github.com/w3c/webappsec-csp/pull/607

closes https://github.com/w3c/webappsec-csp/issues/604
closes https://github.com/w3c/webappsec-permissions-policy/issues/519

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/pull/523.html" title="Last updated on Jul 5, 2023, 2:36 PM UTC (5833bc4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/523/6a673ac...5833bc4.html" title="Last updated on Jul 5, 2023, 2:36 PM UTC (5833bc4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -342,10 +342,13 @@ spec: HEADER-STRUCTURE; urlPrefix: https://httpwg.org/http-extensions/draft-ietf
 
       1. If <var>origin</var> is opaque, return false.
 
+      1. Let <var>url</var> be the result of calling the [=url parser=] on the
+         [=serialization of an origin|serialization=] of <var>origin</var>.
+
       1. Otherwise, [=set/for each=] <a>permissions-source-expression</a> |item| in the <a>allowlist</a>:
 
         1. If the result of running [=Does url match expression in origin with redirect count?=]
-           on a [=url=] of <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
+           on <var>url</var>, |item|, <var>origin</var>, and 0 is true then return true.
 
       1. Return false.
     </div>
--- a/index.bs
+++ b/index.bs
@@ -22,6 +22,8 @@ spec:html; type:element; text:link
 spec:fetch; type:dfn; text:name
 spec:fetch; type:dfn; text:value
 spec:infra; type:dfn; text:list
+spec:csp3; type:dfn; text:scheme-source
+spec:csp3; type:dfn; text:host-source
 </pre>
 <pre class="anchors">
 spec:payment-request; urlPrefix: https://w3c.github.io/payment-request/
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -349,7 +344,7 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
 
       1. Otherwise, [=set/for each=] <a>permissions-source-expression</a> |item| in the <a>allowlist</a>:
 
-        1. If the result of running <a>Does url match expression in origin with redirect count?</a>
+        1. If the result of running [=Does url match expression in origin with redirect count?=]
            on <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
```
clelland, line 348: While you're in here, can you reword this so that we properly cast the first *origin* argument as a URL? You can close out #519 too, if you do that.

index.bs
```diff
@@ -349,8 +344,8 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
 
       1. Otherwise, [=set/for each=] <a>permissions-source-expression</a> |item| in the <a>allowlist</a>:
 
-        1. If the result of running <a>Does url match expression in origin with redirect count?</a>
-           on <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
+        1. If the result of running [=Does url match expression in origin with redirect count?=]
+           on a [=url=] of <var>origin</var>, |item|, <var>origin</var>, and 0 is true then return true.
```
clelland, line 348: "url of origin" isn't a thing, unfortunately -- I think we really need to do something like

1. let |url| be the result of calling the [[=url parser=]](https://url.spec.whatwg.org/#concept-url-parser) on the [[=serialization of an origin|serialization=]](https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin) of |origin|
2. If the result of running ... on |url|, |item|, |origin| and 0...

index.bs
```diff
@@ -351,10 +346,13 @@ spec: CSP; urlPrefix: https://www.w3.org/TR/CSP3/#
 
       1. If <var>origin</var> is an [=opaque origin=], return false.
 
+      1. Let <var>url</var> be the result of calling the [=url parser=] on the
+         [=serialization of an origin|serialization=] of <var>origin</var>.
+
       1. Otherwise, [=set/for each=] <a>permissions-source-expression</a> |item| in the <a>allowlist</a>'s <a>expressions</a>:
```
clelland, line 352: LGTM -- just get rid of the "Otherwise" here, since the previous line isn't the "If" line anymore. (And that line returns anyway, so control flow would never get here if origin was opaque)

clelland (approved):
Looks good with the one comment addressed, thanks!

PR #529, opened 2023-09-11
Add reporting and report-only mode.

clelland:
This is a fairly large change, which adds proper support for reporting and a report-only mode to Permissions Policy. Reporting can be configured for individual features, with a new "report-to" parameter on the header declarations.

The Permissions-Policy-Report-Only header allows a parallel policy to be constructed, which will generate warning reports (with a "Report" disposition) but which will not cause use of the feature to be blocked.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/pull/529.html" title="Last updated on Oct 12, 2023, 7:10 PM UTC (62d9823)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/529/c48a176...62d9823.html" title="Last updated on Oct 12, 2023, 7:10 PM UTC (62d9823)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -229,18 +229,20 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
   </section>
   <section>
     <h3 id="policies">Policies</h3>
-    <p>A <dfn>declared policy</dfn> is a [=struct=] with the following items:</p>
+    <p>A <dfn>declared policy</dfn> is a [=struct=] with the following
+    [=struct/items=]:</p>
 
     <dl dfn-for="declared policy">
       : <dfn>declarations</dfn>
-      :: an [=ordered map=] from [=features=] to [=/allowlists=]
+      :: an [=ordered map=] from [=features=] to [=allowlists=]
 
       : <dfn>reporting configuration</dfn>
       :: an [=ordered map=] from [=features=] to [=strings=]
 
     </dl>
 
-    <p>A <dfn>permissions policy</dfn> is a [=struct=] with the following items:</p>
+    <p>A <dfn>permissions policy</dfn> is a [=struct=] with the following
+    [=struct/items=]:</p>
 
     <dl dfn-for="permissions policy">
       : <dfn>inherited policy</dfn>
@@ -254,7 +256,9 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
     <p>An <dfn export>empty permissions policy</dfn> is a <a>permissions
     policy</a> that has an <a for="permissions policy">inherited policy</a> which
     contains "<code>Enabled</code>" for every <a>supported feature</a>, a <a
-    for="permissions policy">declared policy</a> which is «[],[]».</p>
+    for="permissions policy">declared policy</a> whose [=declared
+    policy/declarations=] and [=declared policy/reporting configuration=] are
+    both empty [=ordered maps=].</p>
   </section>
   <section>
     <h3 id="inherited-policies">Inherited policies</h3>
@@ -305,7 +309,7 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
     <h3 id="policy-directives">Policy directives</h3>
     <p>A <dfn data-lt="policy directive|policy directives">policy
     directive</dfn> is an [=ordered map=], mapping <a>policy-controlled
-    features</a> to corresponding [=/allowlists=] of origins.</p>
+    features</a> to corresponding [=allowlists=] of origins.</p>
     <p>A <a>policy directive</a> is represented in HTTP headers as the
     serialization of an <a>sf-dictionary</a> structure, and in and HTML
     attributes as its ASCII serialization.</p>
@@ -424,8 +428,9 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
     Each Dictionary Member associates a <a>feature</a> with an <a>allowlist</a>.
     The Member Names must be Tokens. If a Member Name is the Token `*`, then it
     will only be used to configure the default reporting endpoint. If a Member
-    Name is not the Token `*`, and also does not name a supported feature, then
-    the Dictionary Member will be ignored by the processing steps.
+    Name is not the Token `*`, and also does not name one of the user agent's
+    [=supported features=], then the Dictionary Member will be ignored by the
+    processing steps.
 
     The Member Values represent <a>allowlists</a>, and must be one of:
     * a String containing the ASCII <a>permissions-source-expression</a>
@@ -446,8 +451,8 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
 <section>
   <h2 id="delivery">Delivery</h2>
   <section>
-    <h3 id="permissions-policy-http-header-field">Permissions-Policy HTTP Header
-    Field</h3>
+    <h3 id="permissions-policy-http-header-field">`Permissions-Policy` HTTP
+    Header Field</h3>
     <p>The &#96;<dfn export http-header
     id="permissions-policy-header"><code>Permissions-Policy</code></dfn>&#96;
     HTTP header field can be used in the [=response=] (server to client) to
@@ -690,22 +695,25 @@ partial interface HTMLIFrameElement {
 
     <p>The {{getAllowlistForFeature(feature)}} method must run the following
     steps:
-    1. Set |result| to an empty list
+    1. Set |result| to an empty list.
     2. Let |origin| be this {{PermissionsPolicy}} object's <a>default
-        origin</a>.
+       origin</a>.
     3. Let |policy| be the <a>observable policy</a> for this
-        {{PermissionsPolicy}} object's <a>associated node</a>.
+       {{PermissionsPolicy}} object's <a>associated node</a>.
     4. If |feature| is not allowed in |policy| for |origin|, return |result|
-    5. Let |allowlist| be |policy|'s declared policy[|feature|]'s [=declared
-        policy/declarations=].
+    5. Let |allowlist| be |policy|'s [=/declared policy=][|feature|]'s
+       [=declared policy/declarations=].
     6. If |allowlist| is the special value `*`:
         1. Append "`*`" to |result|
         2. Return |result|.
     7. If the <a>allowlist</a>'s <a>self-origin</a> is not null,
-       append the <a lt="serialization of an origin">serialization</a> of it to |result|
+       append the <a lt="serialization of an origin">serialization</a> of it to
+       |result|.
     8. If the <a>allowlist</a>'s <a>src-origin</a> is not null,
-       append the <a lt="serialization of an origin">serialization</a> of it to |result|
-    9. Otherwise, for each <a>permissions-source-expression</a> |item| in |allowlist|'s <a>expressions</a>:
+       append the <a lt="serialization of an origin">serialization</a> of it to
+       |result|.
+    9. Otherwise, for each <a>permissions-source-expression</a> |item| in
+       |allowlist|'s <a>expressions</a>:
         1. Append |item| to |result|
     10. Return |result|.
 
@@ -725,7 +733,9 @@ partial interface HTMLIFrameElement {
             2. Set |inherited policy|[|feature|] to |isInherited|.
         4. Return a new <a>permissions policy</a> with <a for="permissions
            policy">inherited policy</a> |inherited policy|, <a
-           for="permissions policy">declared policy</a> «[], []».
+           for="permissions policy">declared policy</a> a [=struct=] with both
+           [=declared policy/declarations=] and [=declared policy/reporting
+           configuration=] new [=ordered maps=].
 
     <p>To get the <dfn>declared origin</dfn> for an Element |node|, run the
     following steps:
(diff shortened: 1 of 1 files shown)


domfarolino (commented):
Haven't made it all the way through yet, but here's a little progress from yesterday's flight.

index.bs
```diff
@@ -786,9 +806,24 @@ partial interface HTMLIFrameElement {
       resulted only in this report being generated (with no further action taken
       by the user agent in response to the violation).
 
-      Note: There is currently no mechanism in place for enabling report-only
-      mode, so [=PermissionsPolicyViolationReportBody/disposition=] will always
-      be set to "enforce".
+  <section>
+    <h3 id="permissions-policy-report-only-http-header-field">Permissions-Policy-Report-Only
+    HTTP Header Field</h3>
+    <p>The &#96;<dfn export http-header
```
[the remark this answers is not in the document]
clelland: Done

index.bs
```diff
@@ -704,8 +724,8 @@ partial interface HTMLIFrameElement {
                 |feature|, |node| and |node|'s <a>declared origin</a>.
             2. Set |inherited policy|[|feature|] to |isInherited|.
         4. Return a new <a>permissions policy</a> with <a for="permissions
-           policy">inherited policy</a> |inherited policy| and <a
-           for="permissions policy">declared policy</a> a new [=ordered map=].
+           policy">inherited policy</a> |inherited policy|, <a
+           for="permissions policy">declared policy</a> «[], []».
```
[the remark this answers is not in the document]
clelland: Done

index.bs
```diff
@@ -229,19 +229,32 @@ spec: RFC8941; urlPrefix: https://datatracker.ietf.org/doc/html/rfc8941#
   </section>
   <section>
     <h3 id="policies">Policies</h3>
+    <p>A <dfn>declared policy</dfn> is a [=struct=] with the following items:</p>
+
+    <dl dfn-for="declared policy">
+      : <dfn>declarations</dfn>
```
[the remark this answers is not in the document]
domfarolino: Eh, tbh this seems fine to me?

index.bs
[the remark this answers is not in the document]
clelland: Wait -- maybe you meant using an actual `&#96` in the header to mean a literal byte string (like in the paragraph below), and not using backticks to get a code font?

clelland:
FYI, after talking with @arichiv, I've removed the special handling for the `*` pseudo-feature for now. We can look at adding it at a later date, or perhaps some different mechanism for setting the default endpoint.

index.bs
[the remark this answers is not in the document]
domfarolino: Hmm, actually I did not mean to actually use `&#96;`. What I had in mind was basically what we do in https://github.com/WICG/fenced-frame/blob/master/spec.bs#L2222-L2223, however using `&#96;` too I think isn't bad?

index.bs
```diff
@@ -671,21 +692,25 @@ partial interface HTMLIFrameElement {
...
     7. If the <a>allowlist</a>'s <a>self-origin</a> is not null,
-       append the <a lt="serialization of an origin">serialization</a> of it to |result|
+       append the <a lt="serialization of an origin">serialization</a> of it to
+       |result|.
     8. If the <a>allowlist</a>'s <a>src-origin</a> is not null,
-       append the <a lt="serialization of an origin">serialization</a> of it to |result|
-    9. Otherwise, for each <a>permissions-source-expression</a> |item| in |allowlist|'s <a>expressions</a>:
+       append the <a lt="serialization of an origin">serialization</a> of it to
+       |result|.
+    9. Otherwise, for each <a>permissions-source-expression</a> |item| in
```
domfarolino, line 712: nit: could be good to link to `[=list/iterate|for each=]` (I think that should work)

index.bs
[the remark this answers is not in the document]
clelland: Gotcha -- but you did mean for the backticks to appear literally in the rendered output.

index.bs
```diff
@@ -786,9 +813,24 @@ partial interface HTMLIFrameElement {
...
+    <h3 id="permissions-policy-report-only-http-header-field">`Permissions-Policy-Report-Only`
+    HTTP Header Field</h3>
+    <p>The &#96;<dfn export http-header
+    id="permissions-policy-report-only-header"><code>Permissions-Policy-Report-Only</code></dfn>&#96;
+    HTTP header field can be used in the [=response=] (server to client) to
+    communicate a <a>permissions policy</a> that should not be enforced by the
+    client, but instead should be used to trigger reports to be sent if any
+    policy declared within it *would* have been violated, had the policy been
+    active.</p>
+    <p><a http-header>Permissions-Policy-Report-Only</a> is a structured header.
```
domfarolino, line 826: nit: My recommendation for all references to headers, (like my above recommendation) would be to wrap these in escaped backticks (or the literal `&#96;`) to get the code formatting. I think that's pretty common in specs, and makes it more clear

index.bs
```diff
@@ -797,12 +839,14 @@ partial interface HTMLIFrameElement {
     ## <dfn export abstract-op id="process-response-policy">Process response policy</dfn> ## {#algo-process-response-policy}
 
     <div class="algorithm" data-algorithm="process-response-policy">
-    Given a [=response=] (|response|) and an [=origin=] (|origin|), this
-    algorithm returns a <a for="permissions policy">declared policy</a>.
+    Given a [=response=] (|response|), an [=origin=] (|origin|), and a boolean
+    (|report-only|), this algorithm returns a [=/declared policy=].
 
+    1. Let |header name| be "<code>Permissions-Policy-Report-Only</code>" if
+      |report-only| is True, or "<code>Permissions-Policy</code>" otherwise.
```
domfarolino, line 846: Lowercase `T` for "true". Same goes for "False" in other instances below. See https://infra.spec.whatwg.org/#booleans. If the spec does this elsewhere, we can maybe take care of this in a separate PR (or not at all, if you prefer, it isn't a big deal really, just a nit).

index.bs
```diff
@@ -797,12 +839,14 @@ partial interface HTMLIFrameElement {
...
     1. Let |parsed header| be the result of executing <a>get a structured
-      field value</a> given "<code>Permissions-Policy</code>" and "dictionary" from
-      |response|’s header list.
+      field value</a> given |header name| and "dictionary" from |response|’s
+      header list.
```
domfarolino, line 849: nit: link to header list here (https://fetch.spec.whatwg.org/#concept-response-header-list) but this could be done in a follow-up since this is pre-existing

index.bs
```diff
@@ -815,13 +859,16 @@ partial interface HTMLIFrameElement {
 
     <div class="algorithm" data-algorithm="construct-policy">
     Given an <a>ordered map</a> (|dictionary|) and an [=origin=] (|origin|), this
-    algorithm will return a <a for="permissions policy">declared policy</a>.
-    1. Let |policy| be an empty [=ordered map=].
-    1. [=map/For each=] |feature-name| → |value| of |dictionary|:
+    algorithm will return a [=/declared policy=].
+    1. Let |declarations| be an empty ordered map.
+    1. Let |reporting-config| be an empty ordered map.
```
domfarolino, line 864: nit: link to ordered map here and above

index.bs
```diff
@@ -815,13 +859,16 @@ partial interface HTMLIFrameElement {
...
+    1. [=map/For each=] |feature-name| → (|value|, |params|) of |dictionary|:
         1. If |feature-name| does not identify any recognized
           <a>policy-controlled feature</a>, then [=iteration/continue=].
         1. Let |feature| be the <a>policy-controlled feature</a> identified by
           |feature-name|.
+        1. If |params|["report-to"] exists, and is a string, then set
```
domfarolino, line 870: Sometimes this spec uses `"report-to"` (code formatting) and other times (like this case) it doesn't. I'd aim to be consistent (probably using code formatting everywhere).

index.bs
```diff
@@ -989,47 +1041,111 @@ partial interface HTMLIFrameElement {
...
+                "<code>Report</code>", and |report-only endpoint|.
+    1. Return result
+
+    </div>
+  </section>
+  <section>
+    ## <dfn abstract-op id="get-reporting-endpoint">Get the reporting endpoint for a feature</dfn> ## {#algo-get-reporting-endpoint}
+
+    <div class="algorithm" data-algorithm="get-reporting-endpoint">
+    Given a feature (|feature|) and a permissions policy (|policy|), this
```
domfarolino, line 1131: Can you link to "permissions policy" here?

index.bs
```diff
@@ -1089,6 +1205,29 @@ partial interface HTMLIFrameElement {
   </section>
 </section>
 
+<section>
+  <h2 id="changes-to-other-specifications">Changes to other specifications</h2>
+
+  <section>
+    <h3 id="changes-to-html">Changes to the HTML specification</h3>
+    Every {{Document}} has a <dfn for="Document">Report-only permissions
```
domfarolino, line 1213: super nit: just make the definition all lowercase

index.bs
domfarolino, line 1131: Link to `[=feature=]`

index.bs
```diff
@@ -989,47 +1041,111 @@ partial interface HTMLIFrameElement {
...
+    </div>
+  </section>
   <section>
     ## <dfn export abstract-op id="report-permissions-policy-violation">Generate report for violation of permissions policy on settings</dfn> ## {#algo-report-permissions-policy-violation}
 
     <div class="algorithm" data-algorithm="report-permissions-policy-violation">
     Given a [=feature=] (|feature|), an <a>environment settings object</a>
-    (|settings|), and an optional string (|group|), this algorithm generates a
-    <a>report</a> about the <a>violation</a> of the policy for |feature|.
+    (|settings|), a String (|disposition|), and an optional string (|group|),
```
domfarolino, line 1146: lowercase "S" for `[=string=]`

domfarolino, line 1146: One problem here is that we always pass in `group`, its just sometimes it is the actual value null. In that case, I think we should either:
 - Adjust the callers to *not* pass it in at all if the string is null (thus taking advantage of the optionality here)
 - Adjust the algorithm to accept a "string-or-null" (commonly seen in HTML Standard) and react to the null case appropriately.

domfarolino, line 1146: The callers of this algorithm seem to pass in as the last argument, an endpoint. But the pre-existing argument here is called `|group|`, which is a confusing name. Maybe there's context here that I'm missing, but I just want to be sure that these are at least corresponding to the same thing here, right? (Just want to double check!)

index.bs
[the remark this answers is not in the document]
domfarolino: Yes!

index.bs
clelland: Will change to `endpoint` here; originally the reporting api dealt with named endpoint groups, where you could configure a number of URLs to try in sequence to deliver a report. That has changed now, and other than in a couple of special cases, we should be naming endpoints instead.

clelland: I'll make it mandatory, and react to null; in earlier iterations of reporting, there was no way to configure the group, and so it was effectively always "default" (we actually never called it from this spec, the intention at the time was to call this from other specs instead, and *maybe* they would choose a different endpoint group).

index.bs
clelland: Will address in a separate cleanup as well. Thanks!

index.bs
```diff
@@ -989,47 +1041,111 @@ partial interface HTMLIFrameElement {
...
   </section>
+  <section>
+    ## <dfn export abstract-op id="is-feature-enabled">Is feature enabled in document for origin?</dfn> ## {#algo-is-feature-enabled}
+
+    <div class="algorithm" data-algorithm="is-feature-enabled">
+    Given a [=feature=] (|feature|), a {{Document}} object
+    (|document|), an [=origin=] (|origin|), and an optional boolean (|report|),
+    with a default value of True, this algorithm returns "<code>Disabled</code>"
+    if |feature| should be considered disabled, and "<code>Enabled</code>"
+    otherwise. If |report| is True, then it will also generate and queue a
```
domfarolino, line 1086: Link to "generate and queue a report"

index.bs
```diff
@@ -1043,23 +1159,18 @@ partial interface HTMLIFrameElement {
...
-        ::  "enforce"
+        ::  |disposition|
 
     1. If the user agent is currently executing script, and can extract the
       source file's URL, line number, and column number from |settings|, then
       set |body|'s [=PermissionsPolicyViolationReportBody/sourceFile=],
       [=PermissionsPolicyViolationReportBody/lineNumber=], and
       [=PermissionsPolicyViolationReportBody/columnNumber=] accordingly.
 
-    1. If |group| is omitted, set |group| to "default".
```
domfarolino, line 1054: Are we ever reacting to it being null here? Maybe I'm misunderstanding https://github.com/w3c/webappsec-permissions-policy/pull/529#discussion_r1357250611, but I expected to see a null check.

clelland: My intention here is actually to put the null check into Reporting -- ideally, even with a null `endpoint`, this *would* generate a report for JavaScript reporting observers, but not actually send it out anywhere over the wire.

Currently Reporting does this indirectly, by creating a report object, notifying observers, and then queuing the report for sending. That report is later discarded when it realizes that there is no valid endpoint with that (null) name.

I filed https://github.com/w3c/reporting/issues/265 to officially support that pattern, which right now works by accident.

index.bs
domfarolino: Addressed below

index.bs
domfarolino: OK, thanks that sounds good!

domfarolino (approved):

PR #545, opened 2024-04-08
Include feature status of some directives

0xedward:
Specifically the following directives:
- `attribution-reporting`
- `capture-surface-control`
- `compute-pressure`
- `direct-sockets`

I found the status of these directives from their respective specs, chromestatus.com and digging through some chromium source code. Let me know if something is incorrect


clelland:
Thanks! I think that the only thing that has changed is that `compute-pressure` has now shipped in Chrome, but I can make that update.

Issue #537, opened 2024-01-29
Send reports for Permissions Policy violations in iframe to parent frame's endpoint


shhnjk:
From https://crbug.com/1501168:
> Permissions Policy violation inside cross-origin iframes are not sent to the top frame's reporting endpoint. While this might be due to a security concern that cross-origin information would be leaked, without this information, it is really difficult to deploy Permissions Policy because sites who's delegating permissions won't know what will break (because they don't receive violation).

So far, the discussion has been leaning to send reports whenever iframe's `allow` attribute was specified with an origin which is blocked by the Permissions Policy.

To implement this, we will likely add [_Is feature enabled in document for origin_](https://w3c.github.io/webappsec-permissions-policy/#is-feature-enabled) to [_Parse policy directive_](https://w3c.github.io/webappsec-permissions-policy/#parse-policy-directive).

We'd like to know if there is a concern to this change.

CC: @arturjanc, @mikewest, @ddworken, @salcho

annevk:
I don't think that addresses the concern raised. And I don't think we can address the concern raised without compromising on the same-origin policy, which we shouldn't do.

shhnjk:
Right, the solution proposed will send a report for each iframes which has an allow attribute with denied origin, but it does not tell whether the permission was used inside an iframe.
I think this is a good compromise, as this does not leak new information to the page (i.e. this can be done through JavaScript), but it is definitely better than the status quo. So I think it's worth implementing.

Any objection to implementing this?

annevk:
Yes, it's not clear to me it's solving an actual problem.

clelland:
@shhnjk, is the idea here that simply constructing an iframe, which tries to delegate a permission that it is not allowed to, would trigger a report in the embedding page?

arturjanc:
Yes, the idea is to report when the document _delegates_ a permission to an iframe whose origin is not allowed to receive the permission in the Permissions Policy configuration. I think this has two main benefits:
1. It doesn't introduce the security concern that @annevk pointed out because the violation doesn't depend on any activity in the iframe itself. The report is sent upon delegation, which is in control of the embedder and thus something the embedder already knows -- it would not expose if the iframe actually uses the delegated permission.
[...]

PR #546, opened 2024-04-30
Send reports for Permissions Policy violations in iframe to parent frame's endpoint

shhnjk:
This change implements permission policy reporting for potential Permissions Policy violations to parent frames.

Currently, Permissions Policy violations inside an iframe is not sent to parent frame, because of security concerns. However, this makes it difficult for websites to roll out Permissions Policy because the site owner does not have insight into permission breakage in iframes.

The change is implemented in a way that it does not leak any new information to parent frame, while sending signals that iframes might be affected by the Permissions Policy.

Fixes: https://github.com/w3c/webappsec-permissions-policy/issues/537

***
<a href="https://pr-preview.s3.amazonaws.com/shhnjk/webappsec-permissions-policy/pull/546.html" title="Last updated on Sep 23, 2024, 8:30 PM UTC (7f6eee5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/546/f15a454...shhnjk:7f6eee5.html" title="Last updated on Sep 23, 2024, 8:30 PM UTC (7f6eee5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1002,7 +1002,8 @@ partial interface HTMLIFrameElement {
         1. Call <a abstract-op>Is feature enabled in document for origin?</a>
            on |feature|, |element|'s [=node document=], |origin|, True, True, and
            |allowAttribute|.
-          Note: The purpose of the call is to generate a report, and not to allow or deny a feature usage.
+
+        Note: The purpose of the call is to generate a report, and not to allow or deny a feature usage.
     1. Return |container policy|.
 
     </div>
--- a/index.bs
+++ b/index.bs
@@ -852,12 +852,9 @@ partial interface HTMLIFrameElement {
       string indicating whether the <a>violated</a> permissions policy was
       enforced in this case.
       [=PermissionsPolicyViolationReportBody/disposition=] will be set to
-      "enforce" if the policy was enforced, "report" if the <a>violation</a>
+      "enforce" if the policy was enforced, or "report" if the <a>violation</a>
       resulted only in this report being generated (with no further action taken
-      by the user agent in response to the violation), "potential-enforce" if the
-      policy was potentially enforced on iframes, or "potential-report" if the
-      potential <a>violation</a> on iframes resulted only in this report being
-      generated.
+      by the user agent in response to the violation).
 
     - <dfn for="PermissionsPolicyViolationReportBody">allowAttribute</dfn>: If
       known, the value of <{iframe}> element's <{iframe/allow}> attribute, or 
@@ -1166,22 +1163,30 @@ partial interface HTMLIFrameElement {
         1. Let |settings| be |document|'s <a>environment settings object</a>.
         1. If |result| is "<code>Disabled</code>":
             1. Let |endpoint| be the result of calling <a abstract-op>Get the
-                reporting endpoint for a feature</a> given |feature| and
-                |policy|.
-            1. Let |disposition| be "<code>potential-enforce</code>" if
-               |potential| is True, or "<code>enforce</code>" otherwise.
-            1. Call <a abstract-op>Generate report for violation of permissions
-                policy on settings</a> given |feature|, |settings|,
-                |disposition|, |endpoint|, and |allowAttribute|.
+               reporting endpoint for a feature</a> given |feature| and
+               |policy|.
+            1. If |potential| is True:
+                1. Call <a abstract-op>Generate report for potential violation
+                   of permissions policy on settings</a> given |feature|,
+                   |settings|, "<code>enforce</code>", |endpoint|, and
+                   |allowAttribute|.
+            1. Else:
+                1. Call <a abstract-op>Generate report for violation of
+                   permissions policy on settings</a> given |feature|, |settings|,
+                   "<code>enforce</code>", and |endpoint|.
         1. Else, if |report-only result| is "<code>Disabled</code>":
             1. Let |report-only endpoint| be the result of calling <a
-                abstract-op>Get the reporting endpoint for a feature</a> given
+               abstract-op>Get the reporting endpoint for a feature</a> given
                |feature| and |report-only policy|.
-            1. Let |disposition| be "<code>potential-report</code>" if
-               |potential| is True, or "<code>report</code>" otherwise.
-            1. Call <a abstract-op>Generate report for violation of permissions
-                policy on settings</a> given |feature|, |settings|,
-                |potential|, |report-only endpoint|, and |allowAttribute|.
+            1. If |potential| is True:
+                1. Call <a abstract-op>Generate report for potential violation of
+                   permissions policy on settings</a> given |feature|, |settings|,
+                   "<code>report</code>", |report-only endpoint|, and
+                   |allowAttribute|.
+            1. Else:
+                1. Call <a abstract-op>Generate report for violation of permissions
+                   policy on settings</a> given |feature|, |settings|,
+                   "<code>report</code>", and |report-only endpoint|.
     1. Return result
 
     </div>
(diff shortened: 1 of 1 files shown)


shhnjk:
Friendly ping @clelland :)

shhnjk:
@clelland, I've introduced `Generate report for potential violation of permissions policy on settings`. PTAL!

index.bs
```diff
@@ -855,6 +856,10 @@ partial interface HTMLIFrameElement {
       resulted only in this report being generated (with no further action taken
       by the user agent in response to the violation).
 
+    - <dfn for="PermissionsPolicyViolationReportBody">allowAttribute</dfn>: If
+      known, the value of <{iframe}> element's <{iframe/allow}> attribute, or
```
clelland, line 860: In what cases would an attribute value not be known? Also, this does not refer to any specific iframe, just the concept of the <iframe> element in the HTML spec.

index.bs
```diff
@@ -980,15 +985,23 @@ partial interface HTMLIFrameElement {
...
+      directive</a> given |allowAttribute|, the [=Document/origin=] of |element|'s
+      [=node document=], and |element|'s <a>declared origin</a>.
     1. If |element|'s <{iframe/allowfullscreen}> attribute is specified, and
        |container policy| does not [=map/contain=] an entry for the
        <code>fullscreen</code> [=policy-controlled feature|feature=].
       1. [=map/Set=] |container policy|[<code>fullscreen</code>] = <a>the
          special value <code>*</code></a>.
+    1. [=map/For each=] [=feature=] in |container policy|:
+      1. If [=feature=]'s <a>allowlist</a> does not [=list/contain=] <a>the special
+         value <code>*</code></a>, then [=list/for each=] [=origins=] in
```
clelland, line 1000: `|origin|` instead of `[=origins=]` here

index.bs
```diff
@@ -980,15 +985,23 @@ partial interface HTMLIFrameElement {
...
-      the [=Document/origin=] of |element|'s [=node document=], and |element|'s
-      <a>declared origin</a>.
+      directive</a> given |allowAttribute|, the [=Document/origin=] of |element|'s
+      [=node document=], and |element|'s <a>declared origin</a>.
     1. If |element|'s <{iframe/allowfullscreen}> attribute is specified, and
        |container policy| does not [=map/contain=] an entry for the
        <code>fullscreen</code> [=policy-controlled feature|feature=].
       1. [=map/Set=] |container policy|[<code>fullscreen</code>] = <a>the
          special value <code>*</code></a>.
+    1. [=map/For each=] [=feature=] in |container policy|:
```
clelland, line 998: `|feature|` instead of `[=feature=]` here and on the next line

index.bs
```diff
@@ -1213,6 +1239,46 @@ partial interface HTMLIFrameElement {
...
+        :   [=PermissionsPolicyViolationReportBody/allowAttribute=]
+        ::  |allowAttribute|
+
+    1. If the user agent is currently executing script, and can extract the
+      source file's URL, line number, and column number from |settings|, then
+      set |body|'s [=PermissionsPolicyViolationReportBody/sourceFile=],
+      [=PermissionsPolicyViolationReportBody/lineNumber=], and
+      [=PermissionsPolicyViolationReportBody/columnNumber=] accordingly.
+
+    1. If |allowAttribute| is not null, then set |body|'s
```
clelland, line 1273: This step isn't necessary; it's already been initialized above.

index.bs
```diff
@@ -980,15 +985,23 @@ partial interface HTMLIFrameElement {
...
     1. If |element|'s <{iframe/allowfullscreen}> attribute is specified, and
        |container policy| does not [=map/contain=] an entry for the
        <code>fullscreen</code> [=policy-controlled feature|feature=].
       1. [=map/Set=] |container policy|[<code>fullscreen</code>] = <a>the
          special value <code>*</code></a>.
+    1. [=map/For each=] [=feature=] in |container policy|:
+      1. If [=feature=]'s <a>allowlist</a> does not [=list/contain=] <a>the special
+         value <code>*</code></a>, then [=list/for each=] [=origins=] in
+         <a>allowlist</a>:
+        1. Executing <a abstract-op>Is feature enabled in document for origin?</a>
```
clelland, line 1002: "Call" or "Execute" rather than "Executing"?

We might want a note here that the algorithm is being called specifically for the side effect of generating a report, not for actual usage of the feature.

index.bs
shhnjk: Done!

index.bs
shhnjk: Done!

index.bs
shhnjk: Done!

index.bs
shhnjk: Done!

index.bs
shhnjk: `PermissionsPolicyViolationReportBody` is used in 2 violation types (i.e. permissions-policy-violation` and `potential-permissions-policy-violation`). And the `allowAttribute` is only set for `potential-permissions-policy-violation` type. Maybe I should say `if defined` insted of `if known`.

For refering to an iframe, what's the best way to indicate the iframe which is in question (in the spec wording)?

clelland: Maybe something like this?

> For reports of potential violations, which can be attributed to a specific <{iframe}> element, the value of the <{iframe/allow}> attribute of that element, or omitted otherwise.

index.bs
```diff
@@ -980,15 +985,25 @@ partial interface HTMLIFrameElement {
...
+      directive</a> given |allowAttribute|, the [=Document/origin=] of |element|'s
+      [=node document=], and |element|'s <a>declared origin</a>.
     1. If |element|'s <{iframe/allowfullscreen}> attribute is specified, and
        |container policy| does not [=map/contain=] an entry for the
        <code>fullscreen</code> [=policy-controlled feature|feature=].
       1. [=map/Set=] |container policy|[<code>fullscreen</code>] = <a>the
          special value <code>*</code></a>.
+    1. [=map/For each=] |feature| in |container policy|:
+      1. If [=feature=]'s <a>allowlist</a> does not [=list/contain=] <a>the special
+         value <code>*</code></a>, then [=list/for each=] |origin| in
```
clelland, line 1000: This might not acutally work, now that we have [Wildcards in Permissions Policy Origins](https://github.com/w3c/webappsec-permissions-policy/pull/516), the allowlist doesn't necessarily contain origins.
Instead, if it's not `*`, it contains [permissions-source-expressions](https://w3c.github.io/webappsec-permissions-policy/#permissions-source-expression) which could be concrete origins, or could include wildcards, like `https://*.github.io` or even `https://*`.
I'm not sure how you'd want to handle that; maybe constructing a "representative origin" for such a pattern, and testing *that*, or ignoring patterns with wildcards (or something else?)

shhnjk:
@clelland, PTAL when you have time!

index.bs
```diff
@@ -1043,17 +1072,19 @@ partial interface HTMLIFrameElement {
     <div class="algorithm"
     data-algorithm="define-inherited-policy-in-container">
     Given a [=policy-controlled feature|feature=] (|feature|), null or a <a>navigable container</a>
-    (|container|), and an <a for="Document">origin</a> for a {{Document}} in
-    that container (|origin|), this algorithm returns the [=inherited policy for
-    a feature|inherited policy value=] for |feature|.
+    (|container|), an <a for="Document">origin</a> for a {{Document}} in
```
clelland, line 1075: The inherited policy for a child document has always been based on the enforced policy in the parent; not the report-only policy of the parent. Report-only policies are local, essentially.

I think what you've done here is correct,  since the default is false, and it's only overridden by the algorithm which tests for potential violations, but I just want to make sure that there's no intention to have report-only policies themselves ever be inherited by child documents.

index.bs
```diff
@@ -764,12 +764,35 @@ partial interface HTMLIFrameElement {
...
     following steps:</p>
         1. Let |inherited policy| be an empty [=ordered map=].
-        3. [=set/For each=] <a>supported feature</a> |feature|:
+        2. [=set/For each=] <a>supported feature</a> |feature|:
             1. Let |isInherited| be the result of running <a abstract-op>Define
                 an inherited policy for feature in container at origin</a> on
                 |feature|, |node| and |node|'s <a>declared origin</a>.
             2. Set |inherited policy|[|feature|] to |isInherited|.
-        4. Return a new <a>permissions policy</a> with <a for="permissions
+            3. If |node| is a <a>navigable container</a>:
```
clelland, line 772: This code is good - it should probably be somewhere else, though, looking at it again. This is a side effect attached to the getter for the observable policy, which would mean that sending reports would be triggered by a script calling `element.featurePolicy,allowedFeaures()` or similar methods.

Is the intention here to have this called when an iframe is created, or when its src/allow attribute changes?

index.bs
```diff
@@ -1212,6 +1189,41 @@ partial interface HTMLIFrameElement {
...
+           policy for feature in container at origin</a> on |feature|,
+           |container| and |container|'s <a>declared origin</a> is
+           "<code>Disabled</code>":
+          1. Let |endpoint| be the result of calling <a abstract-op>Get
+             the reporting endpoint for a feature</a> given |feature| and
+             |document|'s [=Document/permissions policy=].
+          2. Call <a abstract-op>Generate report for potential violation
+             of permissions policy on settings</a> given |feature|,
+             |settings|, "<code>Enforce</code>", and |endpoint|.
+        2. If the result of running <a abstract-op>Define an inherited
```
clelland, line 1212: I think you might want an "Else, " here  - the spec typically only sends one report per check, and the enforcement report pre-empts the report-only report.

clelland:
This is looking really good, @shhnjk  - other than one nit, I think it's probably good to merge

shhnjk:
@clelland feel free to merge this :)

clelland:
Thanks, @shhnjk ! This looks good now.

PR #559, opened 2025-01-31
Add allow and src attribute to potential permissions policy violation report

shhnjk:
In the [intial spec change](https://github.com/w3c/webappsec-permissions-policy/pull/546), I forgot to do a plumbing of `allowAttribute` to potential permissions policy violation report. This change fix that.
Additionally, having `src` attribute will help developers identify which `iframe` is affected, and therefore this chage add that as well.

Bug: #537

***
<a href="https://pr-preview.s3.amazonaws.com/shhnjk/webappsec-permissions-policy/pull/559.html" title="Last updated on Feb 10, 2025, 7:58 PM UTC (1c37e68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/559/8430c86...shhnjk:1c37e68.html" title="Last updated on Feb 10, 2025, 7:58 PM UTC (1c37e68)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -824,6 +824,7 @@ partial interface HTMLIFrameElement {
       readonly attribute long? columnNumber;
       readonly attribute DOMString disposition;
       readonly attribute DOMString? allowAttribute;
+      readonly attribute DOMString? srcAttribute;
     };
   </pre>
 
@@ -860,6 +861,11 @@ partial interface HTMLIFrameElement {
     - <dfn for="PermissionsPolicyViolationReportBody">allowAttribute</dfn>: For
       reports of potential violations, which can be attributed to a specific
       <{iframe}> element, the value of the <{iframe/allow}> attribute of that
+      element, or omitted otherwise.
+
+    - <dfn for="PermissionsPolicyViolationReportBody">allowAttribute</dfn>: For
+      reports of potential violations, which can be attributed to a specific
+      <{iframe}> element, the value of the <{iframe/src}> attribute of that
       element, or omitted otherwise.
 
   <section>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1344,11 +1354,13 @@ partial interface HTMLIFrameElement {
     And in the same section, in step 10, set the new {{Document}}'s
     [=Document/report-only permissions policy=] to |reportOnlyPermissionsPolicy|.
 
-    And in the same section, in step 19 before the return, insert the following step:
+    In <a
```
clelland, line 1357: This implies that this check only applies to iframe elements, and not to any other kind of embed. (<fencedframe>, <frame>, <object>, <embed>, etc.) Is that correct?

index.bs
```diff
@@ -1283,6 +1291,8 @@ partial interface HTMLIFrameElement {
         ::  |disposition|
         :   [=PermissionsPolicyViolationReportBody/allowAttribute=]
         ::  |allowAttribute|
+        :   [=PermissionsPolicyViolationReportBody/allowAttribute=]
```
clelland, line 1294: This should say `[=PermissionsPolicyViolationReportBody/srcAttribute=]`, I believe.

As a separate question, does including it here mean that these keys should exist in the report even if the values are null? Should we try to avoid that?

index.bs
shhnjk: Right, because `allow` attribute is only present in iframe elements.

index.bs
shhnjk: Fixed!
We should avoid setting the key when allow attribute or src attribute is not set. But for potential permissions policy violation, it only triggers when there is a conflict between alllow attribute and permissions policy, so I think it will always be present.

index.bs
```diff
@@ -1281,8 +1289,10 @@ partial interface HTMLIFrameElement {
         ::  null
         :   [=PermissionsPolicyViolationReportBody/disposition=]
         ::  |disposition|
-        :   [=PermissionsPolicyViolationReportBody/allowAttribute=]
+        :   [==]
```
clelland, line 1292: This one disappeared with the latest commit, I think.

index.bs
clelland: Thanks for confirming!

index.bs
clelland: That makes sense - this section does only apply to potential violations.

index.bs
shhnjk: Fixed, sorry about that :(

clelland (approved):
LGTM!

PR #572, opened 2025-07-30
Allow for User Agent discretion when checking permission policy

jkarlin:
The [Selective Permissions Intervention](https://github.com/explainers-by-googlers/selective-permissions-intervention/) allows the UA to Disable features for a given permissions policy check.

While small, this is a significant change in the sense that two calls in the same browsing context can have different responses. Notably, even if `allowedFeatures` says a feature is allowed, it may not be allowed for particular invocations. Any suggestions for additional prose/warnings are welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/jkarlin/webappsec-permissions-policy/pull/572.html" title="Last updated on Mar 3, 2026, 8:33 PM UTC (e3713ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/webappsec-permissions-policy/572/298d386...jkarlin:e3713ba.html" title="Last updated on Mar 3, 2026, 8:33 PM UTC (e3713ba)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1162,6 +1162,7 @@ partial interface HTMLIFrameElement {
     1. Let |report-only result| be the result of calling <a abstract-op>Check
        permissions policy</a>, given |report-only policy|, |feature|, |origin|,
        and |document|'s [=Document/origin=].
+    1. Let |result| be "<code>Disabled</code>" at the user agent's discretion.
     1. If |report| is True:
         1. Let |settings| be |document|'s <a>environment settings object</a>.
         1. If |result| is "<code>Disabled</code>":
--- a/index.bs
+++ b/index.bs
@@ -1162,7 +1162,12 @@ partial interface HTMLIFrameElement {
     1. Let |report-only result| be the result of calling <a abstract-op>Check
        permissions policy</a>, given |report-only policy|, |feature|, |origin|,
        and |document|'s [=Document/origin=].
-    1. Let |result| be "<code>Disabled</code>" at the user agent's discretion.
+    1. Optionally, set |result| to "<code>Disabled</code>" if the user agent deems it necessary.
+
+        Note: The user agent may wish to disallow particular contexts or scripts from using the 
+        API. See the <a 
+        href="https://github.com/explainers-by-googlers/selective-permissions-intervention">
+        Selective Permissions Intervention</a> as an example.
     1. If |report| is True:
         1. Let |settings| be |document|'s <a>environment settings object</a>.
         1. If |result| is "<code>Disabled</code>":
--- a/index.bs
+++ b/index.bs
@@ -1164,10 +1164,10 @@ partial interface HTMLIFrameElement {
        and |document|'s [=Document/origin=].
     1. Optionally, set |result| to "<code>Disabled</code>" if the user agent deems it necessary.
 
-        Note: The user agent may wish to disallow particular contexts or scripts from using the 
-        API. See the <a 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1162,6 +1162,7 @@ partial interface HTMLIFrameElement {
     1. Let |report-only result| be the result of calling <a abstract-op>Check
        permissions policy</a>, given |report-only policy|, |feature|, |origin|,
        and |document|'s [=Document/origin=].
+    1. Let |result| be "<code>Disabled</code>" at the user agent's discretion.
```
domfarolino, line 1165: You want "set" instead of "let", since `|result|` has already been brought into existence on L1159. But I would personally write:

```suggestion
    1. Optionally, set |result| to "<code>Disabled</code>" if the user agent deems it necessary.

       Note: The motivation of letting the user agent intervene here is driven by [... your documentation here describing what happens here in practice, but not setting it in stone, hence the Note ...]
```

jkarlin: Thanks! Done.

index.bs
```diff
@@ -1162,6 +1162,12 @@ partial interface HTMLIFrameElement {
     1. Let |report-only result| be the result of calling <a abstract-op>Check
        permissions policy</a>, given |report-only policy|, |feature|, |origin|,
        and |document|'s [=Document/origin=].
+    1. Optionally, set |result| to "<code>Disabled</code>" if the user agent deems it necessary.
+
+        Note: The user agent may wish to disallow particular contexts or scripts from using the
```
domfarolino, line 1167: I would dedent this by a single space, just leaving the characters aligned with the beginning of "Optionally" above. (I think [`specfmt`](https://github.com/domfarolino/specfmt) should do this for you, if you wanted to play with it.)

domfarolino (approved):
Hey @clelland, what do you think about this?

clelland:
In general, I'm a fan of the idea that the user agent should be the ultimate arbiter of whether a feature is allowed or not, and even if the permissions policy structure is set up perfectly by all parties, the UA may have additional knowledge that a feature should *not* be allowed in a given context.

I like that this phrasing only allows the agent to deny access, and not to arbitrarily grant access as well.

One point about the location of the check; since [Define an inherited policy](https://w3c.github.io/webappsec-permissions-policy/#algo-define-inherited-policy-in-container) doesn't call through to [Is feature enabled](https://w3c.github.io/webappsec-permissions-policy/#algo-is-feature-enabled), this will mean that it's possible for the user agent to intervene like this and disable a feature in one document, but for the same feature to be enabled in one of its embeds. I don't think we've enabled that situation before. I'm not against it, since again, the browser makes that choice, but it might make it easier for sites which are being blocked to find loopholes.

We should also consider whether the (rather neglected) JS API [allowsFeature()](https://w3c.github.io/webappsec-permissions-policy/#dom-permissionspolicy-allowsfeature) and friends should return false in this situation (it's unclear, since that text says "allowed by policy", and doesn't link anywhere useful ☹️)

@annevk for thoughts as well

scheib:
I'm supportive of the patch, but suggest inlining a note sharing the example of protecting against advertisement scripts. This would help the spec be readable with example and not require reader indirection.

jkarlin:
@scheib good idea, I've described the example inline.

jkarlin:
@clelland I added a bit to the allowsFeature API, please take a look.

ghost:
Please allow my third party app access. I'm not receiving messages

PR #583, opened 2026-03-31
Add security/privacy questionnaire for focus-without-user-activation

ffiori:
Add a Security and Privacy questionnaire that serves as self review from the [W3C questionnaire](https://www.w3.org/TR/security-privacy-questionnaire/#questions) that is required for the ChromeStatus security review in order to start an Origin Trial.


security-privacy-questionnaire-focus-without-user-activation
```diff
@@ -0,0 +1,91 @@
...
+
+Not beyond existing web focus behavior. The feature controls whether web content may programmatically move focus in certain embedding scenarios, it does not add new control over the user agent's native UI surfaces.
+
+### 2.13 What temporary identifiers do the features in this specification create or expose to the web?
+
+None.
+
+### 2.14 How does this specification distinguish between behavior in first-party and third-party contexts?
+
+This feature is specifically designed for embedding scenarios and can restrict programmatic focus in cross-origin iframes unless the policy delegates that capability. In other words, it gives embedders control over whether third-party embedded content can move focus away from the embedding context.
```
dandclark: It can restrict it in same-origin frames too, right? Should this just read "can restrict programmatic focus in iframes"?

dandclark: Maybe this should say something like how the policy uses existing permission policy syntax to control whether programmatic focus should be enabled in first party content, third party content, or both.

dandclark (approved):

security-privacy-questionnaire-focus-without-user-activation
ffiori: Thanks for the suggestion, just updated the wording so it's more clear.

dandclark (approved):

ffiori:
Thank you @dandclark for the review.
@clelland could you PTAL?

smilly-boy (approved):
#584

